PR #11, opened 2022-03-19
Dedupe fetch in fetchDataImpl

juni-notifi:
There are some client usage patterns which would result in massive
overfetching. Dedupe if there are pending requests.

Also, there's a high likelihood that clients are using async / await
patterns. To solve for these cases, implement a cache TTL of 1000ms.


juni-notifi:
Cherry-picked in #14

PR #12, opened 2022-03-19
Lower case emails in ensureTarget

juni-notifi:
For emails specifically, we should be case-insensitive to the input.
When sending values to the server, also lower case the emails.

### DIFF
--- a/packages/notifi-react-hooks/lib/utils/ensureTarget.spec.ts
+++ b/packages/notifi-react-hooks/lib/utils/ensureTarget.spec.ts
@@ -130,14 +130,37 @@ describe('ensureEmail', () => {
     id: 'existing',
     emailAddress: 'existing@e.mail',
   } as EmailTarget;
-  let existing = [existingItem];
+
+  const nullItem = {
+    id: null,
+    emailAddress: null,
+  } as EmailTarget;
+
+  let existing = [existingItem, nullItem];
 
   const subject = async (value: string | null) =>
     ensureEmail(service, existing, value);
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/notifi-react-hooks/lib/utils/ensureTarget.spec.ts
```diff
@@ -130,14 +130,37 @@ describe('ensureEmail', () => {
...
+      expect(createSpy).not.toHaveBeenCalled();
+    });
+
+    it('returns the id of the existing target', async () => {
+      const result = await subject('Existing@E.mail');
+      expect(result).toEqual(existingItem.id);
+    });
+
+    it('does not modify the existing array', async () => {
+      await await subject('Existing@E.mail');
```
kargo-k, line 161: nit: `await await`

packages/notifi-react-hooks/lib/utils/ensureTarget.ts
```diff
@@ -12,10 +12,12 @@ export type CreateFunc<Service, T> = (
   value: string,
 ) => Promise<T>;
 export type IdentifyFunc<T> = (arg: T) => string | null;
+export type ValueTransformFunc = (value: string) => string;
 
 const ensureTarget = <Service, T extends Readonly<{ id: string | null }>>(
   create: CreateFunc<Service, T>,
   identify: IdentifyFunc<T>,
+  valueTransform?: ValueTransformFunc,
```
kargo-k, line 20: neat! 💯

kargo-k (approved):
one nit ☝️

PR #20, opened 2022-04-18
Add support for Telegram

nimesh-notifi:


packages/notifi-axios-adapter/lib/fragments/supportedTargetTypesFragment.ts
```diff
@@ -0,0 +1,7 @@
+export const supportedTargetTypesFragment = `
+fragment supportedTargetTypesFragment on SupportedTargetTypesForDapp {
+
```
juni-notifi, line 3: intentionally fetching nothing for now?

packages/notifi-core/lib/operations/GetSupportedTargetTypesForDapp.ts
```diff
@@ -0,0 +1,7 @@
+import { ParameterLessOperation, TargetType } from '../models';
+
+export type GetSupportedTargetTypesResult = ReadonlyArray<TargetType>;
+
+export type GetSupportedTargetTypesForDappService = Readonly<{
+  getSupportedTargetTypesForDapp: ParameterLessOperation<GetSupportedTargetTypesResult>;
```
juni-notifi, line 6: i don't think we should use ParameterLessOperation, we should take a ClientGetSupportedTargetTypesForDappInput as an input right?

nimesh-notifi: yup, I missed that

packages/notifi-axios-adapter/lib/fragments/supportedTargetTypesFragment.ts
nimesh-notifi: nope, definitely not my intention!

nimesh-notifi: query supportedTargets {
  supportedTargetTypesForDapp(getSupportedTargetTypesForDappInput: {
    dappAddress: ""
  }) {
    
  }

PR #38, opened 2022-06-20
Add transaction auth

nimesh-notifi:


package.json
```diff
@@ -38,5 +38,8 @@
     "packages/notifi-node-sample",
     "packages/notifi-react-card",
     "packages/notifi-react-hooks"
-  ]
+  ],
```
alessio-notifi, line 41: This should be a dev dependency

packages/notifi-react-hooks/package.json
```diff
@@ -41,5 +41,6 @@
   "peerDependencies": {
     "react": "^17.0.2 || ^18.0.0",
     "react-dom": "^17.0.2 || ^18.0.0"
-  }
+  },
+  "gitHead": "ecfc53099c4a622add72a955b7dd1c56767fd26a"
```
alessio-notifi, line 45: I wonder why this was added, is this from the alpha release?

alessio-notifi (approved):

PR #42, opened 2022-06-29
MVP-813: Allow users to edit targets when all alerts are disabled

alessio-notifi:
Changes:
* Add `ensureTargetGroup` to `NotifiClient`
* When `subscribe` is called use `ensureTargetGroup` to update targets when all alerts are disabled
* Append `resultKey` to GQL path to improve dev experience when troubleshooting network requests

### DIFF
--- a/packages/notifi-axios-utils/lib/utils/axiosRequest.ts
+++ b/packages/notifi-axios-utils/lib/utils/axiosRequest.ts
@@ -14,10 +14,6 @@ const makeRequestInternal = async <Input, Result>(
   variables: Input,
   config?: AxiosRequestConfig<Input>,
 ): Promise<Result> => {
-  console.log('### test', {
-    encoded: encodeURIComponent(resultKey),
-    resultKey,
-  });
   const { data } = await axiosInstance.post<PostResponse<Result>>(
     `/${encodeURIComponent(resultKey)}`,
     {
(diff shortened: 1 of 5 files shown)


packages/notifi-axios-utils/lib/utils/axiosRequest.ts
```diff
@@ -15,7 +15,7 @@ const makeRequestInternal = async <Input, Result>(
   config?: AxiosRequestConfig<Input>,
 ): Promise<Result> => {
   const { data } = await axiosInstance.post<PostResponse<Result>>(
-    '/',
+    `/${encodeURIComponent(resultKey)}`,
```
nimesh-notifi, line 18: Is this essentially the same as setting x-request-id header? Can we do that instead, or too? We can send that back down in the header as x-response-id

nimesh-notifi (approved):
would be nice to just set the headers for tracing too

packages/notifi-axios-utils/lib/utils/axiosRequest.ts
alessio-notifi: This is just to make it easier to identify GQL requests when troubleshooting:
image

I opened [MVP-822](https://notifi.atlassian.net/browse/MVP-822) and [MVP-822](https://notifi.atlassian.net/browse/MVP-823) to track the work required to support an `x-request-id` header

PR #52, opened 2022-07-21
Add `values` key to filterOptions

juni-notifi:
This key allows us to configure with either an `and` or `or` to evaluate
some values. Each item has a `key` `op` and `value`.

example:

```
const filterOptions = {
  values: {
    and: [
      { key: "foo", op: "eq", value: "A" },
      { key: "price", op: "gt", value: "0" },
      { key: "price", op: "lte", value: "1" },
    ],
  }
}
```

### DIFF
--- a/packages/notifi-core/lib/NotifiClient.ts
+++ b/packages/notifi-core/lib/NotifiClient.ts
@@ -29,11 +29,21 @@ export type AlertFrequency =
   | 'HOURLY'
   | 'DAILY';
 
+export type ValueItemConfig = Readonly<{
+  key: string;
+  op: 'lt' | 'lte' | 'eq' | 'gt' | 'gte';
+  value: string;
+}>;
+
 export type FilterOptions = Partial<{
   alertFrequency: AlertFrequency;
   directMessageType: string;
   threshold: number;
   delayProcessingUntil: string;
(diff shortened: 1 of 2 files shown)


packages/notifi-core/lib/NotifiClient.ts
```diff
@@ -29,11 +29,21 @@ export type AlertFrequency =
   | 'HOURLY'
   | 'DAILY';
 
+export type ValueItemConfig = Readonly<{
+  key: string;
+  op: 'lt' | 'lte' | 'eq' | 'gt' | 'gte';
+  value: string;
```
alessio-notifi, line 35: value should be a number

juni-notifi: i was thinking we'd serialize it to string and let the backend do the explicit conversions to the relevant values to match (based on key, i guess?) -- wdyt? this would let us use things like `equality` on strings

an alternative might be to encode expected type in the API
```
export type ValueItemConfig = Readonly<
  | { key: string, type: 'number', op: NumberOp, value: number }
  | { key: string, type: 'string', op: StringOp, value: string }
    ...
>;
```

alessio-notifi (approved):

PR #56, opened 2022-08-17
Add completeLoginViaTransaction and beginLoginViaTransaction functions in Frontend Client Package

jennifer-notifi:
Ported over `completeLoginViaTransaction` and `beginLoginViaTransaction` functions from `notifi-react-hooks` package into `notifi-frontend-client`

To Do: 
- Remove `walletPublicKey` in configuration and update `dappAddress` name to `tenantName`


packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -1,12 +1,29 @@
 import { NotifiService } from '@notifi-network/notifi-graphql';
-import { UserFragmentFragment } from 'notifi-graphql/lib/gql/generated';
 
+import {
+  BeginLogInByTransactionResult,
+  CompleteLogInByTransactionInput,
+  CompleteLogInByTransactionMutation,
+  UserFragmentFragment,
+  WalletBlockchain,
+} from '../../notifi-graphql/lib/gql/generated';
```
juni-notifi, line 9: we should not use relative paths, since this is a different NPM package

Can you explore why your VS Code isn't resolving via the NPM workspaces?

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -1,12 +1,29 @@
...
 import type { NotifiFrontendConfiguration } from './configuration/NotifiFrontendConfiguration';
 import type { NotifiStorage } from './storage';
 import { notNullOrEmpty } from './utils/notNullOrEmpty';
 
 export type SignMessageFunction = (message: Uint8Array) => Promise<Uint8Array>;
 
+type BeginLoginProps = {
+  walletBlockchain: WalletBlockchain;
+  walletPublicKey: string;
+};
```
juni-notifi, line 19: Omit<BeginLogInByTransactionInput, 'dappAddress'>`

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -1,12 +1,29 @@
...
+
+type CompleteLoginProps = {
+  input: CompleteLogInByTransactionInput;
+};
```
juni-notifi, line 23: `Omit<CompleteLogInByTransactionInput, 'dappAddress' | 'randomUuid'>`

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -1,12 +1,29 @@
...
+
+let clientRandomUuid: string;
```
juni-notifi, line 25: `let clientRandomUuid: string | null = null`

Also, this should be an instance variable, not a module scope variable

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +91,49 @@ export class NotifiFrontendClient {
...
+  async completeLoginViaTransaction({
+    walletBlockchain,
+    walletPublicKey,
+    input,
+  }: BeginLoginProps &
+    CompleteLoginProps): Promise<CompleteLogInByTransactionMutation> {
+    const { transactionSignature } = input;
+    const { dappAddress } = this._configuration;
+
+    if (!clientRandomUuid) {
```
juni-notifi, line 123: i prefer to totally avoid truthiness checks when possible. If we change this to `string | null` we can have an explicit `!== null`

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +91,49 @@ export class NotifiFrontendClient {
...
+      throw 'BeginLoginViaTransaction is required to be called first';
```
juni-notifi, line 124: let's always throw an instance of error: `new Error('...')`

juni-notifi (changes requested):
Looks great! Just a few points of feedback:
- Please remove `.DS_Store` (add it to .gitignore?)
- In general, we should expose as much of the underlying graphql input params as we can
- For the SDK, we need much stricter type safety than we do in the Notifi Hub.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Ah, it does actually, just didn't realize that it was a best practice :)

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Sweet, +1 to utility types: typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

jennifer-notifi:
> Looks great! Just a few points of feedback:
> 
> * Please remove `.DS_Store` (add it to .gitignore?)
> * In general, we should expose as much of the underlying graphql input params as we can
> * For the SDK, we need much stricter type safety than we do in the Notifi Hub.

Thank you for the comments! I removed `DS_Store` using these steps: https://gist.github.com/lohenyumnam/2b127b9c3d1435dc12a33613c44e6308 and added it to the `.gitignore` file. I also reviewed what the DS_Store file: https://buildthis.com/ds_store-files-and-why-you-should-know-about-them/.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -16,6 +29,7 @@ export class NotifiFrontendClient {
     private _configuration: NotifiFrontendConfiguration,
     private _service: NotifiService,
     private _storage: NotifiStorage,
+    private clientRandomUuid: string | null,
```
juni-notifi, line 32: initialize the value? `= null`

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +88,49 @@ export class NotifiFrontendClient {
...
+
+    const result = await this._service.beginLogInByTransaction({
+      walletAddress: walletAddress,
+      walletBlockchain: walletBlockchain,
+      dappAddress,
+    });
+
+    const nonce = result.beginLogInByTransaction.nonce;
+
+    this.clientRandomUuid = await window.crypto.randomUUID();
```
juni-notifi, line 106: oooh ideally we can pull in a crypto dependency rather than depending on `window`

I know this is what happens in the current repo but it does feel wonky (this function will only be successful in a browser? i guess?)

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +88,49 @@ export class NotifiFrontendClient {
...
+  }: BeginLoginProps): Promise<BeginLogInByTransactionResult> {
+    const { dappAddress } = this._configuration;
+
+    const result = await this._service.beginLogInByTransaction({
+      walletAddress: walletAddress,
+      walletBlockchain: walletBlockchain,
+      dappAddress,
+    });
+
+    const nonce = result.beginLogInByTransaction.nonce;
```
juni-notifi, line 104: these objects are nullable, but i don't see null-safe access

Can you share your typescript configuration with me? I think something's not right in the IDE

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +88,49 @@ export class NotifiFrontendClient {
...
+  }
+
+  async completeLoginViaTransaction({
+    walletBlockchain,
+    walletAddress,
+    transactionSignature,
+  }: CompleteLoginProps): Promise<CompleteLogInByTransactionMutation> {
+    const { dappAddress } = this._configuration;
+
+    if (this.clientRandomUuid == null) {
```
juni-notifi, line 118: prefer explicit equivalence `===`

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +88,49 @@ export class NotifiFrontendClient {
...
+      throw new Error(
+        'BeginLoginViaTransaction is required to be called first',
+      );
+    }
+
+    const result = await this._service.completeLogInByTransaction({
+      walletAddress: walletAddress,
+      walletBlockchain: walletBlockchain,
+      dappAddress,
+      randomUuid: this.clientRandomUuid,
```
juni-notifi, line 128: it's not an issue here, but in general, for async functions, it's good practice to store mutable variables in enclosing scope into a local `const` so the same value is used throughout the function's execution (between yields)

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Sorry I missed this, will update

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Ah, yes, explicit would be better

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Sorry for the miss. Will initialize.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: During our call, we confirmed the TS config is correct and the IDE is not advising on implementing chaining since the object cannot be nullable.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: We mentioned during our call that it is best to call `const clientRandomUuid = this._clientRandomUuid;` and use `clientRandomUuid` throughout the function so no one can update it before the other parts of the function fire off.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: After research on this, we concluded via Slack DM that using `crypto` alone does not rely on the browser being open: https://developer.mozilla.org/en-US/docs/Web/API/crypto_property

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +81,61 @@ export class NotifiFrontendClient {
...
+      const data = encoder.encode(nonce + ruuid);
+
+      const hashBuffer = await crypto.subtle.digest('SHA-256', data);
+
+      const hashArray = Array.from(new Uint8Array(hashBuffer));
+      const logValue =
+        'Notifi Auth: 0x' +
+        hashArray.map((b) => b.toString(16).padStart(2, '0')).join('');
+      return { nonce: logValue };
+    } else throw new Error('Failed to begin login process');
```
juni-notifi, line 112: nit: prefer early return pattern to avoid deep indentation 

```
if (nonce === null) {
  throw ...
}

...
```

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -73,4 +81,61 @@ export class NotifiFrontendClient {
...
+  }
+
+  async completeLoginViaTransaction({
+    walletBlockchain,
+    walletAddress,
+    transactionSignature,
+  }: CompleteLoginProps): Promise<Types.CompleteLogInByTransactionMutation> {
+    const { dappAddress } = this._configuration;
+    const clientRandomUuid = this._clientRandomUuid;
```
juni-notifi, line 121: after this assignment, I think we should clear out `this._clientRandomUuid` -- what do you think?

juni-notifi (approved):
Nice, looks good to me.

Stupid question, but `npx lerna run build` completes with no errors, right? We should probably put that on a pre-commit hook somewhere.

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: good call - so the code doesn't have to find it at the end

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Yes, I think that clearing it out makes sense in case `beginViaTransaction` has to run again in the same SDK instance.

```
const clientRandomUuid = this._clientRandomUuid;

this._clientRandomUuid = null;
```

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: > Nice, looks good to me.
> 
> Stupid question, but `npx lerna run build` completes with no errors, right? We should probably put that on a pre-commit hook somewhere.

Screen Shot 2022-08-18 at 11 16 39 AM

Yes, it does! Let me look into adding the command in a pre-commit and how that would work.

PR #57, opened 2022-08-17
Remove walletPublicKey in Frontend Configuration

jennifer-notifi:


packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
```diff
@@ -13,6 +13,16 @@ import { notNullOrEmpty } from './utils/notNullOrEmpty';
...
+type LoginProps = {
+  walletPublicKey: string;
+  readonly signMessage: SignMessageFunction;
+};
+
+type SignMessageProps = {
+  walletPublicKey: string;
+  readonly signMessage: SignMessageFunction;
+  readonly timestamp: number;
+};
```
juni-notifi, line 25: why can't these be fully Readonly?

```
type LoginProps = Readonly<{
  ...
}>
```

juni-notifi (commented):
I think you can easily revert the docs via `git checkout` -- let me know if you need help with it.

I'll provide a final review after we get the begin-login branch merged! (please change this PR's base against sdk-v2 at that time)

packages/notifi-frontend-client/lib/NotifiFrontendClient.ts
jennifer-notifi: Good point here, yes, walletAddress should not be changes as well

jennifer-notifi:
> I think you can easily revert the docs via `git checkout` -- let me know if you need help with it.
> 
> I'll provide a final review after we get the begin-login branch merged! (please change this PR's base against sdk-v2 at that time)

Will give it a try, will put this in draft mode for now as I will probably have merge conflicts/need to rebase.

jennifer-notifi:
> I think you can easily revert the docs via `git checkout` -- let me know if you need help with it.
> 
> I'll provide a final review after we get the begin-login branch merged! (please change this PR's base against sdk-v2 at that time)

Sweet, I found good steps here: https://stackoverflow.com/questions/39459467/remove-a-modified-file-from-pull-request. I have to be more explicit with my searches on code :)

packages/notifi-frontend-client/lib/storage/LocalForageStorageDriver.ts
```diff
@@ -26,9 +26,8 @@ const getEnvPrefix = (env: NotifiEnvironment): string => {
 export const createLocalForageStorageDriver = ({
   env,
   dappAddress,
-  walletPublicKey,
 }: NotifiFrontendConfiguration): StorageDriver => {
-  const keyPrefix = `${getEnvPrefix(env)}:${dappAddress}:${walletPublicKey}`;
```
juni-notifi, line 31: ah unfortunately we need to somehow preserve the publickey in the key.

probably means we need to expose a second parameter

PR #73, opened 2022-10-06
creates input separators

Nathan-Notifi:
Darkmode:
image

Lightmode:
image

Differing text:
image

Missing label:
image

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
@@ -14,8 +14,8 @@ import { EventTypeDirectPushRow } from './EventTypeDirectPushRow';
 import type { EventTypeUnsupportedRowProps } from './EventTypeUnsupportedRow';
 import { EventTypeUnsupportedRow } from './EventTypeUnsupportedRow';
 import {
-  NotifiInputFooters,
   NotifiInputLabels,
+  NotifiInputSeparators,
 } from './NotifiSubscriptionCard';
 
 export type SubscriptionCardV1Props = Readonly<{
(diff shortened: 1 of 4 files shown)


packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
```diff
@@ -49,6 +52,23 @@ export const SubscriptionCardV1: React.FC<SubscriptionCardV1Props> = ({
...
+          <div
+            className={clsx(
+              'NotifiInputSeparator__content',
+              inputSeparators.first.classNames?.content,
+            )}
+          >
+            {inputSeparators?.first?.content}
+          </div>
+        </div>
+      ) : null}
```
juni-notifi, line 71: hmmm, having `inputSeparator.first / second` makes it a little unwieldy to use i think -- if I have email disabled, for example, where should I expect "first" separator to go? 

I'd probably rather see some concept of "email.footer / sms.footer / telegram.footer" instead, because that makes it feel more related to each input field (and makes it clear where the "separator" will go

juni-notifi (changes requested):

packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
Nathan-Notifi: ooo, great point. Let me update that

Nathan-Notifi: updated it so it looks like this.
i might improve this to also handle errors or something.
image

nubstick27 (approved):
nit: prooobably want to consolidate naming (including classnames) to either "separators" or "footers"

juni-notifi (approved):

PR #84, opened 2022-10-18
adds explicit box sizing property

Nathan-Notifi:


packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -82,6 +82,10 @@
   max-height: 22px;
 }
 
+.NotifiEmailInput__container {
+  box-sizing: content-box;
+}
+
```
juni-notifi, line 88: can you put it on the shared styles above? maybe on line 83

juni-notifi (approved):

PR #86, opened 2022-10-18
updates subscribe method

Nathan-Notifi:
https://user-images.githubusercontent.com/105258726/196400654-e848c436-eaf8-445c-9572-8230815c85bb.mov

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
@@ -1,19 +1,9 @@
 import type { Alert } from '@notifi-network/notifi-core';
 import { PropsWithChildren } from 'react';
-import React, {
-  createContext,
-  useCallback,
-  useContext,
-  useRef,
-  useState,
-} from 'react';
+import React, { createContext, useContext, useState } from 'react';
 
-import {
-  EventTypeConfig,
-  FetchedCardView,
-  useFetchedCardState,
-} from '../hooks';
-import type { AlertConfiguration } from '../utils';
+import { FetchedCardView, useFetchedCardState } from '../hooks';
+
 import { NotifiParams } from './NotifiContext';
 
 export type NotifiSubscriptionData = Readonly<{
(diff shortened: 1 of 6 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -279,37 +299,38 @@ export const useNotifiSubscribe: () => Readonly<{
...
+
+      for (let i = 0; i < alerts.length; i++) {
+        const existingAlert = alerts[i];
+        console.log('existing alert', existingAlert);
         if (
-          source === undefined ||
-          source.id === null ||
-          filter === undefined ||
-          filter.id === null
+          existingAlert !== undefined &&
```
Nathan-Notifi, line 323: it gets funky here. i create the alerts for configurations but if not, i attempt to update all existing alerts

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -279,37 +299,38 @@ export const useNotifiSubscribe: () => Readonly<{
           );
         }
 
+        const alert = await client.createAlert({
+          emailAddress: finalEmail,
+          filterId: filter?.id ?? '',
```
Nathan-Notifi, line 304: how should i handle these?

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -279,37 +299,38 @@ export const useNotifiSubscribe: () => Readonly<{
...
+          targetGroupName: 'Default',
+          telegramId: finalTelegramId,
+        });
+
+        newResults[currentConfigurationData.alertName] = alert;
+      }
+    } else {
+      const alerts = data.alerts;
+
+      for (let i = 0; i < alerts.length; i++) {
```
juni-notifi, line 319: this part is not actually sufficient

when we have alerts, we need to:
- identify which alertConfigurations have a corresponding alert
  - if the corresponding alert is the same (i.e. same source, same target) do not do anything
  - if the corresponding alert is not the same, delete it and recreate it using configuration
- identify which alerts have a `null` alertConfiguration and delete them
- if there exists an alert whose name doesn't appear in alertConfiguration, leave them alone

happy to talk through what i mean

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -58,6 +64,7 @@ export const NotifiSubscriptionContextProvider: React.FC<
   PropsWithChildren<NotifiParams>
 > = ({ children, ...params }) => {
   const [email, setEmail] = useState<string>('');
+  const [eventTypes, setEventTypes] = useState<EventTypeConfig>([]);
```
juni-notifi, line 67: oh hm let's not leak this all the way down

possible to call setAlertConfiguration instead from the FetchedCard? using eventTypes mapped to alert configurations

juni-notifi (commented):
some quick thoughts, lmk

packages/notifi-react-card/lib/utils/AlertConfiguration.ts
```diff
@@ -1,7 +1,10 @@
+// import { AlertConfiguration } from './AlertConfiguration';
```
juni-notifi, line 1: ?

juni-notifi (approved):

PR #93, opened 2022-10-25
PR-1472 - [3/3] - Alert History Pagination 

jennifer-notifi:
Demo Video (Alert History with Virtual Scroll and confirm pagination through Network calls - Please note I was setting the MESSAGES_PER_PAGE constant to 8 in this video for demo purposes):

https://user-images.githubusercontent.com/108922885/198110163-f79c584d-c05a-4662-b0b3-9635bb8072a7.mov

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
@@ -4,7 +4,7 @@ import {
   NotificationHistoryEntry,
 } from '@notifi-network/notifi-core';
 import clsx from 'clsx';
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { ListRange, Virtuoso } from 'react-virtuoso';
 
 import {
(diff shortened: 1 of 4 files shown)


packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -68,15 +70,84 @@ export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
...
+        )}
+        itemContent={(index, notification) => {
+          setCurrentIndex(index);
+          return (
+            <AlertCard key={notification.id} notification={notification} />
+          );
+        }}
+      />
+    );
+  }, [alertHistoryData]);
```
juni-notifi, line 150: wouldn't this just be a component that takes setIsScrolling, setVisibleRange, setCurrentIndex, and allNodes as props?

it's tricky because the callback re-renders on alertHistoryData, but actually has a dependency on allNodes.

I think memoized render functions are a little harder to maintain, so my suggestion is to extract a separate component.

juni-notifi (changes requested):

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
jennifer-notifi: Nice, catch! I will componetize the virtual container and update the dependency of allNodes (I must have been playing around with both as data at one point).

jennifer-notifi: I reviewed the code more closely and I realized that I will not need a useCallback for this because I am now appending the paginated node data into `allNodes`. 

I originally was trying to implement a `useCallback` as a workaround to render data at that index prior to having an array with all of the paginated data. I have reverted it back.

Here is the video of the testing to confirm that this works:

https://user-images.githubusercontent.com/108922885/198160510-fe6b4e71-93dc-4c96-9f52-ef0586ee2e97.mov

Please let me know if you have any other feedback!

juni-notifi (approved):

PR #94, opened 2022-10-26
Small fixes

juni-notifi:
- Added react-virtuoso dependency
- Renamed kebab-case properties to camelCase for Reactisms


jennifer-notifi (approved):
Good catch on the camelcase of the SVG properties and dependency!

PR #102, opened 2022-11-10
[MVP-1522] Intercom notification setting view

marukohao:
As a user I can click on settings in the chat window, which lets me edit and see the verification status of my email address, phone number, telegram ID, and browser alert selection.

Test:
updated
Screen Shot 2022-11-11 at 9 33 37 AM

Untitled

Story:
https://notifi.atlassian.net/browse/MVP-1522

### DIFF
--- a/packages/notifi-react-card/lib/assets/BackArrow copy.tsx
+++ b/packages/notifi-react-card/lib/assets/BackArrow copy.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+
+export type Props = Readonly<{
+  className?: string;
+}>;
+
+export const BackArrow: React.FC<Props> = ({ className }: Props) => {
+  return (
+    <svg
+      width="16"
+      height="16"
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <path
+        d="M16 7H3.83L9.42 1.41L8 0L0 8L8 16L9.41 14.59L3.83 9H16V7Z"
+        fill="#80829D"
+      />
+    </svg>
+  );
+};
--- a/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
@@ -10,6 +10,7 @@ export type NotifiEmailInputProps = Readonly<{
     input: string;
     label: string;
     errorMessage: string;
+    button: string;
   }>;
   copy?: DeepPartialReadonly<{
     placeholder: string;
@@ -18,6 +19,7 @@ export type NotifiEmailInputProps = Readonly<{
   disabled: boolean;
   intercomEmailInputStyle?: string;
   intercomEmailInputContainerStyle?: string;
+  intercomView?: boolean;
 }>;
 
 export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
@@ -26,9 +28,15 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   disabled,
   intercomEmailInputStyle,
   intercomEmailInputContainerStyle,
+  intercomView,
 }: NotifiEmailInputProps) => {
-  const { email, setEmail, setEmailErrorMessage, emailErrorMessage } =
-    useNotifiSubscriptionContext();
+  const {
+    intercomCardView,
+    email,
+    setEmail,
+    setEmailErrorMessage,
+    emailErrorMessage,
+  } = useNotifiSubscriptionContext();
 
   const validateEmail = () => {
     if (email === '') {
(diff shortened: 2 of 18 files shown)


packages/notifi-react-card/lib/assets/BackArrow copy.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
jennifer-notifi, line 1: I think that this file does not need to be committed - I think that you are using the original back arrow file?

packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
```diff
@@ -122,6 +124,37 @@ export const IntercomCard: React.FC<
...
+          <div className={'NotifiIntercomCardSettingContent__container'}>
+            <div
+              className={clsx(
+                'NotifiIntercomCard__subtitle2',
+                classNames?.subtitle2,
+              )}
+            >
+              {companySupportDescription}
+            </div>
+            <NotifiIntercomFTUNotificationTargetSection
```
jennifer-notifi, line 141: Can you please add the icons in each target? (ex. Mail icon, cell phone icon, Telegram icon, etc.)

Screen Shot 2022-11-11 at 9 47 51 AM

packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
```diff
@@ -47,9 +55,22 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
 
   return (
     <>
-      <label className={clsx('NotifiEmailInput__label', classNames?.label)}>
-        {copy?.label}
-      </label>
+      {intercomView ? (
+        intercomCardView.state === 'settingView' ? (
+          <div
```
jennifer-notifi, line 60: Nit: This Resend verification link will appear if in the data, the destination has not been verified. You can take steps to start implementing the logic by passing in a isVerified data and/or setting a setState.

packages/notifi-react-card/lib/components/NotifiTelegramInput.tsx
```diff
@@ -25,14 +27,29 @@ export const NotifiTelegramInput: React.FC<NotifiTelegramInputProps> = ({
...
+    useNotifiSubscriptionContext();
 
   return (
     <>
-      <label className={clsx('NotifiSmsInput__label', classNames?.label)}>
-        {copy?.label}
-      </label>
+      {intercomView ? (
+        intercomCardView.state === 'settingView' ? (
+          <div
```
jennifer-notifi, line 39: Similar here as well.

Nit: This Resend verification link will appear if in the data, the destination has not been verified. You can take steps to start implementing the logic by passing in a isVerified data and/or setting a setState.

jennifer-notifi (changes requested):
Overall, the components are starting to look really good. Some nit and polish suggestions :)

packages/notifi-react-card/lib/assets/BackArrow copy.tsx
marukohao: as the back arrow icons are not exactly the same(same, color, weight), I renamed the back arrow as IntercomBackArrow.

packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
marukohao: attached the updated screenshot in the description box.

packages/notifi-react-card/lib/assets/BackArrow copy.tsx
jennifer-notifi: We could pass in a different color and size but yes, I know that could be hard.

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -104,6 +105,9 @@ export const useNotifiSubscribe: () => Readonly<{
       setPhoneNumber(phoneNumber ?? '');
 
       const telegramTarget = targetGroup?.telegramTargets[0];
+      if (telegramTarget !== null && emailTarget?.isConfirmed === false) {
```
jennifer-notifi, line 108: Ah, I think you are looking for `if (telegramTarget !== null && telegramTarget?.isConfirmed === false)`.

Actually, a `telegramConfirmationUrl` should not return back if the telegram Id has been confirmed. Instead, in the `NotifiTelegramInput`, you can expose the `telegramConfirmationUrl` and check if it is null or not.

marukohao: double checked with the code in notifi hub, yeah, you are right, we can use telegramConfirmationUrl to check if the telegram has been confirmed or not.

jennifer-notifi (approved):
Thank you for the updates! It looks good from a UI standpoint, I'm sure we'll make more updates when we add the data APIs :)

PR #105, opened 2022-11-15
Alert History Empty State Default Card Size / Shows With Empty Nodes

jennifer-notifi:
- If nodes does not exist OR length is zero, show empty card state
- Subscription Card default height size is 450px, width is 300px
- Fix - If no telegram or text message, do not show an extra separator

Configuration with no alert history (and two out of three inputs):

Screen Shot 2022-11-16 at 10 49 57 AM

Screen Shot 2022-11-16 at 10 50 52 AM

Screen Shot 2022-11-16 at 10 51 07 AM

Configuration with alert history (and all inputs):

Screen Shot 2022-11-16 at 10 54 43 AM

Screen Shot 2022-11-16 at 10 55 33 AM

Screen Shot 2022-11-16 at 2 25 51 PM

Alert History Icons and Date Padding preserved even on narrow screens:
Screen Shot 2022-11-18 at 9 43 12 AM

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/defaults.css
+++ b/packages/notifi-react-card/lib/components/subscription/defaults.css
@@ -87,7 +87,7 @@
   justify-content: space-between;
   width: 100%;
   overflow: hidden;
-  margin: 20px 0;
+  padding: 15px 0;
 }
 
 .NotifiAlertHistory__content {
--- a/packages/notifi-react-card/lib/components/subscription/defaults.css
+++ b/packages/notifi-react-card/lib/components/subscription/defaults.css
@@ -39,6 +39,8 @@
   background-color: var(--notifi-card-background);
   border-radius: 5px;
   padding: 20px 20px 5px 20px;
+  height: 600px;
+  justify-content: space-between;
 }
 
 .NotifiSubscriptionCard__container *:not(:last-child) {
@@ -51,6 +53,7 @@
   align-items: baseline;
   font-size: 12px;
   line-height: 14px;
+  flex-grow: 1;
 }
 
 .NotifiFooter__logoSvg {
@@ -103,6 +106,11 @@
   border-radius: 5px;
 }
 
+.NotifiAlertHistory__headerSection {
+  display: flex;
+  flex-direction: column;
+}
+
 .NotifiAlertHistory__header {
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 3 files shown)


packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -149,6 +149,8 @@
   line-height: 14px;
   color: var(--notifi-secondary-font-color);
   text-align: center;
+  height: 400px;
+  line-height: 400px;
```
juni-notifi, line 153: I think we shouldn't use line-height to try and center the text.

Also, the alert description shouldn't take up a height of 400px, instead, the base card should have a default height defined and this component should grow to fill the  space.

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -148,7 +148,7 @@ export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
           classNames?.dividerLine,
         )}
       />
-      {alertHistoryData?.nodes ? (
+      {alertHistoryData?.nodes && alertHistoryData?.nodes?.length > 0 ? (
```
juni-notifi, line 151: i'd prefer to see this pulled out so we don't have an implicit conversion

e.g.: on line 128, add `const nodeLength = alertHIstoryData?.nodes?.length ?? 0;` and here, use `nodeLength > 0`

juni-notifi (changes requested):
Would you mind making the card have a fixed height and using flex-grow to size the contents instead?

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
jennifer-notifi: Thank you for letting me know about implicit conversions.

I learned that implicit conversion happens when expression contains operands of different built-in types, and no explicit casts are present, the compiler converts one of the operands so that the types match. 

The reason why `const nodeLength = alertHIstoryData?.nodes?.length ?? 0`; will not run into implicit conversion issues is because it will return 0 if `alertHistoryData?.nodes?.length` is `null `or `undefined `or `alertHistoryData?.nodes?.length value` if it is not null or undefined. 

In other words, it will always return a number.

Nathan-Notifi (approved):

jennifer-notifi:
Apologies, due to the prioritization of Acala wallets, I forgot about this PR. I will review and make the updates and/or open a different PR by EOD Monday.

Nathan-Notifi:
good to know@ looks like there are some conflicts so i'd rebase as well.

jennifer-notifi:
Opening a new PR

PR #107, opened 2022-11-17
[MVP-1599] - [2/2] render of Intercom conversation messages - without polling

marukohao:
hard coding(used existing conversation id to get one conversation thread) a conversation messages
updated logics in useIntercomChat to get the correct data structure to render messages
implement pagination

hide sender name and sender avatar component until we have the correct field exposed in the future.
need to ask Sam with a design of loading message state but using wording "Loading..." for now as the state.
will implement polling or apollo subscriptions in another PR.

Test:
Untitled

Story:
https://notifi.atlassian.net/browse/MVP-1599

### DIFF
--- a/packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
@@ -1,17 +1,24 @@
 import clsx from 'clsx';
 import React from 'react';
 
+import {
+  formatConversationDateTimestamp,
+  formatConversationStartTimestamp,
+} from '../../utils/datetimeUtils';
+
 export type ChatMessageDateProps = Readonly<{
   classNames?: Readonly<{
     container: string;
     content: string;
   }>;
-  date?: string;
+  createdDate: string;
+  isStartDate?: boolean;
 }>;
 
 export const ChatMessageDate: React.FC<ChatMessageDateProps> = ({
   classNames,
-  date = 'August 1',
+  createdDate,
+  isStartDate = true,
 }) => {
   return (
     <div
@@ -26,7 +33,9 @@ export const ChatMessageDate: React.FC<ChatMessageDateProps> = ({
           classNames?.content,
         )}
       >
-        {date}
+        {isStartDate
+          ? formatConversationStartTimestamp(createdDate)
+          : formatConversationDateTimestamp(createdDate)}
       </div>
     </div>
   );
--- a/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
@@ -1,40 +1,96 @@
-import clsx from 'clsx';
 import React from 'react';
+import { Virtuoso } from 'react-virtuoso';
 
-import { ChatMessageDate, ChatMessageDateProps } from './ChatMessageData';
+import { useIntercomChat } from '../../hooks/useIntercomChat';
+import { ChatMessageDate, ChatMessageDateProps } from './ChatMessageDate';
+import {
+  ChatWindowIntroSection,
+  ChatWindowIntroSectionProps,
+} from './ChatWindowIntroSection';
 import { MessageList, MessageListProps } from './MessageList';
 
 export type ChatMessageSectionProps = Readonly<{
   classNames?: Readonly<{
-    container: string;
-    content: string;
-    date: ChatMessageDateProps['classNames'];
+    chatWindowIntro: ChatWindowIntroSectionProps['classNames'];
     messageList: MessageListProps['classNames'];
+    date: ChatMessageDateProps['classNames'];
   }>;
   chatMessageSectionIntroContent?: string;
 }>;
 
 export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
   classNames,
-  chatMessageSectionIntroContent = 'What can we help you with today?',
+  chatMessageSectionIntroContent,
 }) => {
+  const {
+    conversation,
+    setIsScrolling,
+    setVisibleRange,
+    setAtTop,
+    isLoading,
+    hasNextPage,
+  } = useIntercomChat({
+    conversationId: '8c5ccf41d93c4bb9b5e1e1014002e923',
+  });
+
   return (
-    <div
-      className={clsx(
-        'NotifiIntercomChatMessageSection__container',
-        classNames?.container,
+    <>
+      {conversation.feed.length === 0 ? (
+        <ChatWindowIntroSection
+          classNames={classNames?.chatWindowIntro}
+          startDate={conversation.createdDate}
+          chatMessageSectionIntroContent={chatMessageSectionIntroContent}
+        />
+      ) : null}
+
+      {conversation.feed.length === 0 ? null : (
+        <Virtuoso
+          atTopStateChange={setAtTop}
+          className="virtual-container"
+          data={conversation.feed}
+          followOutput="auto"
+          isScrolling={setIsScrolling}
+          rangeChanged={setVisibleRange}
+          itemContent={(index, feed) => {
+            return (
+              <div key={index}>
+                {isLoading ? (
+                  <div
+                    style={{
+                      display: 'flex',
+                      justifyContent: 'center',
+                      padding: '10px',
+                    }}
+                  >
+                    Loading...
+                  </div>
+                ) : null}
+                {index === 0 && !hasNextPage && (
+                  <ChatWindowIntroSection
+                    classNames={classNames?.chatWindowIntro}
+                    startDate={conversation.createdDate}
+                    chatMessageSectionIntroContent={
+                      chatMessageSectionIntroContent
+                    }
+                    inVirtualContainerStyle={
+                      'ChatWindowIntro__virtualContainer'
+                    }
+                  />
+                )}
+                {index != 0 && (
+                  <ChatMessageDate
+                    classNames={classNames?.date}
+                    createdDate={feed.timestamp}
+                    isStartDate={false}
+                  />
+                )}
+                <MessageList classNames={classNames?.messageList} feed={feed} />
+              </div>
+            );
+          }}
+          style={{ width: '364px', height: '290px' }}
+        />
       )}
-    >
-      <ChatMessageDate classNames={classNames?.date} />
-      <div
-        className={clsx(
-          'NotifiIntercomChatMessageSectionIntro__content',
-          classNames?.content,
-        )}
-      >
-        {chatMessageSectionIntroContent}
-      </div>
-      <MessageList classNames={classNames?.messageList} />
-    </div>
+    </>
   );
 };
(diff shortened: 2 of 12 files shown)


packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
```diff
@@ -26,7 +33,9 @@ export const ChatMessageDate: React.FC<ChatMessageDateProps> = ({
           classNames?.content,
         )}
       >
-        {date}
+        {isStartDate
+          ? formatConversationStartTimestamp(createdDate)
+          : formatConversationDateTimestamp(createdDate)}
```
jennifer-notifi, line 38: Can you please check on this? It looks like you are showing the formatted date, no matter what?

packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
```diff
@@ -43,12 +39,13 @@ export const MessageGroup: React.FC<MessageGroupProps> = ({
               : 'NotifiIntercomChatOutgoingMessage__container',
             classNames?.messageContainer,
           )}
+          key={index}
         >
-          {isIncoming ? (
+          {/* {isIncoming ? (
```
jennifer-notifi, line 44: Remove comment

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -1,40 +1,96 @@
...
+    conversationId: '8c5ccf41d93c4bb9b5e1e1014002e923',
+  });
+
   return (
-    <div
-      className={clsx(
-        'NotifiIntercomChatMessageSection__container',
-        classNames?.container,
+    <>
+      {conversation.feed.length === 0 ? (
```
jennifer-notifi, line 38: Could you possibly do `conversation.feed.length === 0 ? <ChatWindowIntroSection /> : <Virtuoso />`?

I am just thinking about readability.

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -1,40 +1,96 @@
...
+          atTopStateChange={setAtTop}
+          className="virtual-container"
+          data={conversation.feed}
+          followOutput="auto"
+          isScrolling={setIsScrolling}
+          rangeChanged={setVisibleRange}
+          itemContent={(index, feed) => {
+            return (
+              <div key={index}>
+                {isLoading ? (
```
jennifer-notifi, line 57: Can you advise what is the intended use of the loading state? When should we see the loading state?

packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
```diff
@@ -65,17 +62,17 @@ export const MessageGroup: React.FC<MessageGroupProps> = ({
                   classNames?.sender,
                 )}
               >
-                {message.sender}
+                {/* {message.sender} */}
```
jennifer-notifi, line 65: Should we still show the sender?

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -1,40 +1,96 @@
...
+                    style={{
+                      display: 'flex',
+                      justifyContent: 'center',
+                      padding: '10px',
+                    }}
+                  >
+                    Loading...
+                  </div>
+                ) : null}
+                {index === 0 && !hasNextPage && (
```
jennifer-notifi, line 68: Should the Chat Window Intro Section always appear?

If it is at zero, it should always appear, no matter if it has a next page. The next page value in the API data tells us if there is more data than the pre-defined amount that we intended to display.

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -1,40 +1,96 @@
...
+                    startDate={conversation.createdDate}
+                    chatMessageSectionIntroContent={
+                      chatMessageSectionIntroContent
+                    }
+                    inVirtualContainerStyle={
+                      'ChatWindowIntro__virtualContainer'
+                    }
+                  />
+                )}
+                {index != 0 && (
```
jennifer-notifi, line 80: How are you showing the date?

Though it is not visible in the demo, I wonder if this may cause the date to render more than needed. I interpret this as if the index is anything other than at the state, render a date.

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -1,40 +1,96 @@
...
+                    classNames={classNames?.date}
+                    createdDate={feed.timestamp}
+                    isStartDate={false}
+                  />
+                )}
+                <MessageList classNames={classNames?.messageList} feed={feed} />
+              </div>
+            );
+          }}
+          style={{ width: '364px', height: '290px' }}
```
jennifer-notifi, line 91: Can we allow passing in a variable to override these?

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -1,32 +1,148 @@
...
-  const getConversationMessages = useCallback(async () => {
-    const response = await client.getConversationMessages(input);
-    return response;
+  // initialization - load first batch of messages
+  useEffect(() => {
+    getConversationMessages();
   }, []);
 
+  // load more messages when scrolling to the top
+  useEffect(() => {
```
jennifer-notifi, line 86: Ah, I think that you are already appending the messages as the user scrolls to `ChatMessages` state which is an array, correct? If they scroll up, they should be able to render what is in `ChatMessages`? 

I am not sure if need to call `getConversationMessages` again as the user scrolls up.

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -1,32 +1,148 @@
...
+    if (hasNextPage && atTop && isScrolling && visibleRange.startIndex === 0) {
+      getConversationMessages();
+    }
+  }, [hasNextPage, atTop, isScrolling, visibleRange.startIndex]);
+
+  const conversation = useMemo(() => {
```
jennifer-notifi, line 92: Can you please explain what is happening with `conversation`?

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -1,32 +1,148 @@
-import { useCallback } from 'react';
+import { ConversationMessagesEntry } from '@notifi-network/notifi-core/dist/models/ConversationMessages';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { ListRange } from 'react-virtuoso';
 
 import { useNotifiClientContext } from '../context';
+import { formatConversationDateTimestamp } from '../utils/datetimeUtils';
+
+export type ChatConversation = FeedEntry &
```
jennifer-notifi, line 8: Nit: If you are not using the type anywhere else but in the file, you can remove 'export' (because you don't need to export it to anywhere else).

If you have time, it would be great if you could check for the other types that you created for intercom as well :)

jennifer-notifi (changes requested):

packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
marukohao: Oh, this is the sender's avatar that we may need to render when we have the correct data field to exposed, so comment this for now, mentioned in the description box.

packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
marukohao: same as the sender's avatar

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
marukohao: So the conversation is the information we need to render the data, like message direction and messages feed, all the logics are adding here for getting the needed data(followed the pattern of xMS chat in hub).

packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
marukohao: Do you mean we need to show 'today' when it's today? The logic has been included in the `formatConversationStartTimestamp`

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
marukohao: Yeah, we should only show intro section when this is the very first message, but as we are fetching more messages each time the user scrolling, the first message always has been changed . But we don't want to show the intro section with the new fetched first message until it is the  actually very first message in the entire conversation list.

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
marukohao: We need to call getConversationMessages each time the user scrolls up to fetch more message data.

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
marukohao: the date has been rendered for each feed, we don't need the date to be rendered when index is equal to 0, cause we have the created conversation date in the intro section.

packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
jennifer-notifi: check in the client if there is another api that we need to pull that includes the user's avatar

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
jennifer-notifi: Jennifer is working on the classNames in Alert History, will let you know

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
jennifer-notifi: if it scrolls up, then we will get the data. if it scrolls down, then we grab the data

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
jennifer-notifi: Can you please implement a scroll seek? https://virtuoso.dev/scroll-seek-placeholders/

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
jennifer-notifi: Thank you. As we chatted in our huddle, I would suggest that you rename the variable and also put the if statement condition in a more readable `const`. A comment would be helpful as well!

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
jennifer-notifi: Ah I think you are saying that if we are index === 0 and it's the last page of messages. 

Could we also make this clear? perhaps we can do something like `const IsFirstIndexOnLastPage = index === 0 && !hasNextPage`?

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
jennifer-notifi: We agreed that we will implement a scroll seek later. Hao has created a ticket in the backlog.

jennifer-notifi (approved):

PR #108, opened 2022-11-20
[MVP-1630 ] Send Message Implement

marukohao:
When use hit the send button, firing ConversationMessageMutationCreate query to update conversation messages.

Test:
Untitled

Story:
https://notifi.atlassian.net/browse/MVP-1630

### DIFF
--- a/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
@@ -53,6 +53,7 @@ export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
         />
       ) : (
         <Virtuoso
+          //TODO: improvement to add the scrollSeekConfiguration property
           atTopStateChange={setAtTop}
           className="virtual-container"
           data={conversation.feed}
--- a/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
@@ -8,12 +8,17 @@ import {
   ChatWindowIntroSectionProps,
 } from './ChatWindowIntroSection';
 import { MessageList, MessageListProps } from './MessageList';
+import {
+  SendMessageSection,
+  SendMessageSectionProps,
+} from './SendMessageSection';
 
 export type ChatMessageSectionProps = Readonly<{
   classNames?: Readonly<{
     chatWindowIntro: ChatWindowIntroSectionProps['classNames'];
     messageList: MessageListProps['classNames'];
     date: ChatMessageDateProps['classNames'];
+    sendMessageSection: SendMessageSectionProps['classNames'];
   }>;
   chatMessageSectionIntroContent?: string;
   messageListWidth?: string;
@@ -33,8 +38,9 @@ export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
     setAtTop,
     isLoading,
     hasNextPage,
+    sendConversationMessages,
   } = useIntercomChat({
-    conversationId: '8c5ccf41d93c4bb9b5e1e1014002e923',
+    conversationId: 'f03c78b1ffee4d7eb2ff030f6e665cba',
   });
 
   return (
@@ -47,8 +53,6 @@ export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
         />
       ) : (
         <Virtuoso
-          //TODO: improvement to add the scrollSeekConfiguration property
-          //to render a placeholder element instead of the actual item if the user scrolls too fast
           atTopStateChange={setAtTop}
           className="virtual-container"
           data={conversation.feed}
@@ -57,7 +61,6 @@ export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
           rangeChanged={setVisibleRange}
           itemContent={(index, feed) => {
             const isFirstIndexOnLastPage = index === 0 && !hasNextPage;
-
             return (
               <div key={index}>
                 {isLoading ? (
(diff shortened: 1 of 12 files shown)


packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
```diff
@@ -47,8 +53,6 @@ export const ChatMessageSection: React.FC<ChatMessageSectionProps> = ({
         />
       ) : (
         <Virtuoso
-          //TODO: improvement to add the scrollSeekConfiguration property
```
jennifer-notifi, line 50: Can we please still add this to-do? I think that it is still important

jennifer-notifi, line 50: Can we please still add this to-do? I think that it is still a good to-do :)

packages/notifi-react-card/lib/components/intercom/SendMessageSection.tsx
```diff
@@ -4,16 +4,23 @@ import React, { useState } from 'react';
...
     button: string;
   }>;
+  sendConversationMessages: (message: string) => void;
 }>;
 
 export const SendMessageSection: React.FC<SendMessageSectionProps> = ({
   classNames,
+  sendConversationMessages,
 }) => {
   const [sendMessage, setSendMessage] = useState<string>('');
```
jennifer-notifi, line 17: Could we update the initial state to `null` instead? 

The sendConversationMessage function will fire if the message is `''`. We do disable the Send button if the message is `''` but it may be good extra precautions to default `sendMessage` to `null`.

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -27,15 +27,17 @@ export type FeedEntry = {
...
   first?: number;
+  userId?: string;
   conversationId: string;
 }>;
 
 export const useIntercomChat = ({
   first = 5,
   conversationId,
-}: GetConversationMessagesInputProps) => {
+  userId = '0ff9528b-9bd4-444f-adcb-482673b31c96',
```
jennifer-notifi, line 39: Hmm, would it be cleaner to pass the parameters that are not used in every single function in the respective functions that are using it instead of top level? 

For example, `userId` and first are only being used by the `getConversationMessages` function. So should we only ask the parameter values at the function level instead at the hook level?

`const getConversationMessages = useCallback((first, userId))`

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -80,55 +84,72 @@ export const useIntercomChat = ({
...
       createdDate:
         chatMessages?.[chatMessages.length - 1]?.createdDate ??
         new Date().toISOString(),
       lastMessage: chatMessages?.[0],
     };
   }, [chatMessages, conversationId]);
 
+  const sendConversationMessages = useCallback(
+    (message: string) => {
+      if (message) {
```
jennifer-notifi, line 137: In this case, I think that you will always get a message because message always has to be a string and if not optional. 

In this case, I would recommend returning early if the message is not what you wanted. 

Perhaps if message == '' return or throw some error

However, this message may change if you consider my feedback above.

jennifer-notifi (commented):
Overall a pretty good start! Some feedback that I'm open to opinions on :)

packages/notifi-react-card/lib/components/intercom/ChatMessageSection.tsx
marukohao: Ah, delete the to-do accidentally, added

packages/notifi-react-card/lib/components/intercom/SendMessageSection.tsx
marukohao: Could you explain why the default value null for sendMessage is better than empty string, yeah, as you mentioned, we have made the condition that we disabled the button when it's an empty string. And if we made the default value to null, the type for sendMessage should be string | null instead of just string.

jennifer-notifi: Yes, I think overall, we need to have a better pre-cautions to prevent accidental sending of an empty message. In addition, to making the send button disabled, we should also enforce it in the function as well.

I do think the `null` message is better because it is more explicit than an empty string.

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
marukohao: Yeah, I agreed we could pass the param directly in the function. but as the params we have(conversationId, first, userId) will always be the same value during the entire flow, it will be more convenient for us to pass the params in when using the useIntercomChat hook, and we don't need to pass the value every time when calling the function. For example, if we need to call getConversationMessages twice from the useIntercomChat, and we didn't have the params passed in the top level, we have to pass the param every time that we call getConversationMessages.

jennifer-notifi: Hmm, I think conversationId and userId is probably okay since it is intuitive to a dev that it is on the top level since it pertains to the conversation as a whole and as you mentioned, might be variables that are used quite regularly.

Maybe we can still pass `first` an optional parameter at the function level (that defaults to 5 if nothing is passed in) because it is very specific to `getConversationMessages` and I don't see it as a parameter that will need changed.

Actually, what is even better is to have a default number at the query level. That way, you can make parameter optional and not have to pass in a default value at the component level because you already assign a default at the query level. 

Something like:

```
const QUERY = `query conversationMessages(
  $getConversationMessagesInput: GetConversationMessagesInput!
  $after: String
  $first: Int = 5
```

marukohao: updated to pass first as an optional parameter at the function level(getConversationMessage).

jennifer-notifi (approved):
Thank you for being open to the feedback and for reasoning things out :)

PR #109, opened 2022-11-21
Mvp 1602/sdk example

zain-notifi:


zain-notifi:
will update the readme later today

packages/notifi-react-example/public/robots.txt
```diff
@@ -0,0 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
```
Nathan-Notifi, line 1: What does this do?

zain-notifi: comes default with create-react-app. I think it helps block web crawlers. We can remove if unnecessary!

Nathan-Notifi: i think that's fine!

juni-notifi (approved):
Nice, LGTM

I kinda wish we had gone with React 18 instead, but we can take that on incrementally.

For folks using this: Remember to run `npx lerna run build` for the example repo to be able to pick up your local changes!

PR #111, opened 2022-11-22
[MVP-1649] implement polling for intercom chat

marukohao:
implement polling for intercom chat
update intercom chat flow, when users come back to use intercom, bring them directly to the chat view instead of to the start chat view.

Test:
Untitled1

Story:
https://notifi.atlassian.net/browse/MVP-1649

### DIFF
--- a/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
@@ -32,7 +32,6 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   intercomEmailInputStyle,
   intercomEmailInputContainerStyle,
   intercomView,
-  hasChatAlert = false,
 }: NotifiEmailInputProps) => {
   const {
     email,
@@ -40,6 +39,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
     setEmailErrorMessage,
     emailErrorMessage,
     emailIdThatNeedsConfirmation,
+    intercomCardView,
   } = useNotifiSubscriptionContext();
 
   const { resendEmailVerificationLink } = useNotifiSubscribe({
@@ -68,7 +68,9 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   return (
     <>
       {intercomView ? (
-        email && hasChatAlert && emailIdThatNeedsConfirmation != '' ? (
+        email &&
+        intercomCardView.state === 'settingView' &&
+        emailIdThatNeedsConfirmation != '' ? (
           <div
             onClick={handleClick}
             className={clsx(
--- a/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
@@ -68,7 +68,6 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   return (
     <>
       {intercomView ? (
-        email &&
         intercomCardView.state === 'settingView' &&
         emailIdThatNeedsConfirmation != '' ? (
           <div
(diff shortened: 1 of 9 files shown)


packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
```diff
@@ -58,6 +58,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   return (
     <>
       {intercomView ? (
+        email &&
```
juni-notifi, line 61: why is this necessary?

packages/notifi-react-card/lib/components/NotifiTelegramInput.tsx
```diff
@@ -40,6 +40,7 @@ export const NotifiTelegramInput: React.FC<NotifiTelegramInputProps> = ({
   return (
     <>
       {intercomView ? (
+        telegramId &&
```
juni-notifi, line 43: why is this necessary?

packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
```diff
@@ -155,6 +173,9 @@ export const IntercomCard: React.FC<
         </>
       );
       break;
+    case 'loadingView':
+      view = null;
```
juni-notifi, line 177: would be good to have a loading state somehow! maybe as a follow-up

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -52,6 +52,32 @@ export const useIntercomChat = ({
...
+            if (nodes[0].id === chatMessages[0].id) {
+              return;
+            } else {
+              nodes.forEach((node, index) => {
+                if (node.id === chatMessages[0].id) {
+                  const newMessages = nodes.slice(0, index);
+                  setChatMessages([...newMessages, ...chatMessages]);
+                }
+              });
+            }
```
juni-notifi, line 74: it's probably better to always add all the messages and dedupe on `id`, sorted by message created time.

This logic seems reliant on all new messages being "newer" than the last known newest message, which feels like it could run into edge cases with client-side insert of sendMessage results.

juni-notifi (commented):
lmk if you'd like to talk about the algorithm for merging messages!

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
marukohao: I am adding the chatMessages as dependency here to avoid the edge case for sending messages, but yeah, it's still insecure. Let me redo the logic as you suggested 👍

packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
marukohao: I was thinking of if there is no email address showing here, we don't need to show the verification link, hmm, we could probably just use the emailIdThatNeedsConfirmation to check it. Let me remove the email here.

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -52,6 +55,35 @@ export const useIntercomChat = ({
...
+        .getConversationMessages({
+          first: 50,
+          getConversationMessagesInput: { conversationId },
+        })
+        .then((response) => {
+          if (Array.isArray(response.nodes)) {
+            const nodes = response.nodes;
+            if (nodes[0].id === chatMessages[0].id) {
+              return;
+            } else {
```
juni-notifi, line 70: actually i don't think we need to have this conditional

packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -52,6 +55,35 @@ export const useIntercomChat = ({
...
+              const chatMessageIds = chatMessages.map((message) => message.id);
+              const dedupeNewMessages = nodes.filter(
+                (node) => !chatMessageIds.includes(node.id),
+              );
```
juni-notifi, line 74: it would be nice if chatMessageIds were a Set instead (using `reduce` instead of `map`) to improve runtime of `includes` check

juni-notifi (approved):
thanks for the update!

PR #113, opened 2022-11-23
adds updates for alert history experience

Nathan-Notifi:
This isn't in the finished state but having some updates.

The event renderers will be updated so i'd ignore all of them except for `Broadcast`

# Alert Details
image

# On hover
image


juni-notifi (approved):
Code LGTM

PR #114, opened 2022-11-23
[MVP-1643] - connect create support intercom chat with client

marukohao:
set up createSupport mutation and add conversationParticipant field in getConversationMessage query
hook up createSupport mutation in client

Test:
Untitled1

Story:
https://notifi.atlassian.net/browse/MVP-1643

### DIFF
--- a/packages/notifi-axios-adapter/lib/mutations/createSupportImpl.ts
+++ b/packages/notifi-axios-adapter/lib/mutations/createSupportImpl.ts
@@ -1,0 +1,37 @@
+import {
+  collectDependencies,
+  makeRequest,
+} from '@notifi-network/notifi-axios-utils';
+import { DeleteTargetGroupResult } from '@notifi-network/notifi-core';
+
+const DEPENDENCIES: string[] = [];
+
+const MUTATION = `
+mutation createSupport {
+  createSupportConversation {
+    id
+    conversationType
+    conversationGates {
+       id
+    }
+    name
+    createdDate
+     participants {
+      conversationParticipantType
+       profile {
+         id
+          preferredAddress
+           preferredName
+       }
+       resolvedName
+     }
+  }
+}
+`.trim();
+
+const createSupportImpl = makeRequest<createSupportResult>(
+  collectDependencies(...DEPENDENCIES, MUTATION),
+  'createSupport',
+);
+
+export default createSupportImpl;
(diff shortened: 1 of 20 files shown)


packages/notifi-axios-adapter/lib/mutations/createSupportImpl.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  collectDependencies,
+  makeParameterLessRequest,
+} from '@notifi-network/notifi-axios-utils';
+import { CreateSupportResult } from '@notifi-network/notifi-core';
+
+const DEPENDENCIES: string[] = [];
+
+const MUTATION = `
+mutation createSupport {
```
jennifer-notifi, line 10: Nit: Could we update the mutation name to `createSupportConversation` and cross the function / file names? 

It would help with readability and understanding what the implementation does, and most, if not, mutations in the `axios-adapter` follows this convention. It should work because `mutation` whatevername is the name of the function that assess the `createSupportConversation` mutation in the backend.

If you are under a time crunch, no worries, keep it like this and I can approve.

packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -359,6 +359,11 @@
   background-color: var(--notifi-placeholder-color);
 }
 
+.NotifiIntercomSendMessageSection__button:disabled
+{
```
jennifer-notifi, line 363: Can you please put `NotifiIntercomSendMessageSection__button:disabled {` on one line?

packages/notifi-axios-adapter/lib/mutations/createSupportImpl.ts
marukohao: updated the function name in useNotifiClient from createSupport to createSupportConversation

jennifer-notifi: Hey Hao, thank you for the update! I was wondering if you could do it for the components AND the mutation name as well. I'll approve because it is a nit but you can get back to it when you can.

jennifer-notifi (approved):
Note to nit: We should update the mutation, components, and file names to `createSupportConversation` for clarity. I will approve, given Hao can get to it at some point.

packages/notifi-axios-adapter/lib/mutations/createSupportImpl.ts
marukohao: Hi Jennifer, thanks for pointing out, I have updated the mutation name and all the name from createSupport to createdSupportConversation.

PR #117, opened 2022-11-27
Move icons to TSX files

juni-notifi:
We don't have a separate SVG pre-parser, so simply embed the contents into TSX files.

Seems like vector.svg was unused.


Nathan-Notifi (approved):
Ah great to know!

Thanks, @juni-notifi

PR #118, opened 2022-11-28
[MVP-1638] fix day bubble in chat only needs to show once per day

marukohao:
day bubble showed in the chat for each conversation feed
add isSameDate params to check on showing date or not for each feed

test:
Untitled11

story:
https://notifi.atlassian.net/browse/MVP-1638

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useIntercomChat.ts
+++ b/packages/notifi-react-card/lib/hooks/useIntercomChat.ts
@@ -150,7 +150,7 @@ export const useIntercomChat = ({
     const feed = messageGroups.map((messageGroup, index): FeedEntry => {
       const firstMessage = messageGroup[0];
 
-      const sameDate = () => {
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/hooks/useIntercomChat.ts
```diff
@@ -146,10 +147,24 @@ export const useIntercomChat = ({
       }
     });
 
-    const feed = messageGroups.map((messageGroup): FeedEntry => {
+    const feed = messageGroups.map((messageGroup, index): FeedEntry => {
       const firstMessage = messageGroup[0];
 
+      const sameDate = () => {
```
Nathan-Notifi, line 153: can we rename this to `isSameDate` since it's returning a boolean?

Nathan-Notifi (approved):

PR #120, opened 2022-11-29
Implement Acala Login

jennifer-notifi:
Testing in `notifi-react-example` with login function:

https://user-images.githubusercontent.com/108922885/204475849-5cda8c89-2f89-4dad-83be-38bfef09a550.mov


packages/notifi-core/lib/NotifiClient.ts
```diff
@@ -222,6 +222,13 @@ export type AptosSignMessageFunction = (
   nonce: number,
 ) => Promise<string>;
 
+type hexString = `0x${string}`;
+
+export type AcalaSignMessageFunction = (
+  acalaAddress: string,
```
juni-notifi, line 228: is the client responsible for providing this value? When would it be different from the `walletPublicKey` that's provided through the AcalaParams?

jennifer-notifi: Ah, yes, `acalaAddress` is already being passed in on the Context level. You're right, I actually don't need this.

I was adding `acalaAddress` to make it more explicit that the admin client does need to provide it to Acala's `signMessage` function.

However, I already use the acalaAddress from the config so I don't need this:

```
const { accountAddress } = params;

 const message = `${SIGNING_MESSAGE}${accountAddress}${dappAddress}${timestamp.toString()}`;
      const messageBuffer = stringToHex(message);

      const signedBuffer = await signer.signMessage(
        accountAddress,
        messageBuffer,
      );
```

jennifer-notifi: @juni-notifi checking in to see if this resolves your question!

juni-notifi (approved):

PR #121, opened 2022-11-29
brings react card into react example repo

Nathan-Notifi:
image

Updates the `react-example` to use the react card for development. We had a lengthier method to build and test; hopefully this should ease some pain in developing on the sdk card.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31500,7 +31500,6 @@
       "dependencies": {
         "@ethersproject/bytes": "^5.7.0",
         "@manahippo/aptos-wallet-adapter": "^1.0.2",
-        "@notifi-network/notifi-react-hooks": "^0.42.1",
         "@solana/wallet-adapter-base": "^0.9.18",
         "@solana/wallet-adapter-react": "^0.15.24",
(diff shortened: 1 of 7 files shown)


juni-notifi (approved):
How come you didn't have to change package.json inside the `notifi-react-example`?

If the card repo was already there, let's remove the react-hooks dependency in this commit.

Nathan-Notifi:
Oh actually, receiving an error:
```
errors
: 
[{message: "The current user is not authorized to access this resource.",…}]
0
: 
{message: "The current user is not authorized to access this resource.",…}
extensions
: 
{code: "AUTH_NOT_AUTHENTICATED"}
locations
: 
[{line: 99, column: 3}]
message
: 
"The current user is not authorized to access this resource."
path
: 
["notificationHistory"]
```

It appears that an auth token isn't being sent for notification history until after submitting the user login info. any thoughts?

juni-notifi:
we can fix the notificationHistory issue separately.

Just a heads up @marukohao this PR will probably conflict with yours (#126 )

PR #123, opened 2022-11-30
[MVP-1687] fix create chat alert for intercom card

marukohao:
Test:
Untitled (1)

Story:
https://notifi.atlassian.net/browse/MVP-1687

### DIFF
--- a/packages/notifi-react-card/lib/utils/AlertConfiguration.ts
+++ b/packages/notifi-react-card/lib/utils/AlertConfiguration.ts
@@ -24,7 +24,7 @@ export const chatConfiguration = (): AlertConfiguration => {
     sourceType: 'NOTIFI_CHAT',
     createSource: {
       address: '*',
-    }
+    },
   };
 };
 
--- a/packages/notifi-react-card/lib/utils/AlertConfiguration.ts
+++ b/packages/notifi-react-card/lib/utils/AlertConfiguration.ts
@@ -22,6 +22,9 @@ export const chatConfiguration = (): AlertConfiguration => {
       alertFrequency: 'THREE_MINUTES',
     },
     sourceType: 'NOTIFI_CHAT',
+    createSource: {
+      address: '*',
+    }
   };
 };
 


jennifer-notifi (commented):
Good catch on this! 

Can you please double check if it should be `address`?

I think it should be `blockchainAddress: *`  since the `CreateSourceResult` type has `blockchainAddress` instead of `address`

You can use `packages/notifi-core/lib/operations/CreateSource.ts` and then check `CreateSourceResult`.

Let me know!

marukohao:
> Good catch on this!
> 
> Can you please double check if it should be `address`?
> 
> I think it should be `blockchainAddress: *` since the `CreateSourceResult` type has `blockchainAddress` instead of `address`
> 
> You can use `packages/notifi-core/lib/operations/CreateSource.ts` and then check `CreateSourceResult`.
> 
> Let me know!

Thanks for the review, I have tested locally, source and alert can be created successfully, the `address` here is just for the chatConfiguration, when create the alert, it will be pass in as the value for blockchainAddress.
```
            source = await client.createSource({
              blockchainAddress: createSourceParam.address,
              name: createSourceParam.address,
              type: sourceType,
            });
```

jennifer-notifi:
> > Good catch on this!
> > Can you please double check if it should be `address`?
> > I think it should be `blockchainAddress: *` since the `CreateSourceResult` type has `blockchainAddress` instead of `address`
> > You can use `packages/notifi-core/lib/operations/CreateSource.ts` and then check `CreateSourceResult`.
> > Let me know!
> 
> Thanks for the review, I have tested locally, source and alert can be created successfully, the `address` here is just for the chatConfiguration, when create the alert, it will be pass in as the value for blockchainAddress.
> 
> ```
>             source = await client.createSource({
>               blockchainAddress: createSourceParam.address,
>               name: createSourceParam.address,
>               type: sourceType,
>             });
> ```

Thank you for the clarification! Apologies here, configurations are set type `AlertConfiguration` which are typed as `createSource: { address }`

packages/notifi-react-card/lib/utils/AlertConfiguration.ts
```diff
@@ -22,6 +22,9 @@ export const chatConfiguration = (): AlertConfiguration => {
       alertFrequency: 'THREE_MINUTES',
     },
     sourceType: 'NOTIFI_CHAT',
+    createSource: {
+      address: '*',
+    }
```
jennifer-notifi, line 27: Nit: Can you please add a comma at the end of createSource object?

```
createSource: {
      address: '*',
},
```
I received an eslint error alert so if you should check if eslint is working for you. :)

jennifer-notifi (approved):
Approved, considering that you will handle the nit below

PR #124, opened 2022-11-30
[MVP-1691] fix chat window header layout

marukohao:
fix chat window header layout
changed className in chat window header component to match in css

add onKeyDown in textarea, user can send message by hitting enter key.

Screenshot 2022-11-29 at 7 45 32 PM
before:

after:
Screenshot 2022-11-29 at 8 01 48 PM

### DIFF
--- a/packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
@@ -23,18 +23,12 @@ export const ChatWindowHeader: React.FC<ChatWindowHeaderProps> = ({
   };
   return (
     <div
-      className={clsx(
-        'NotifiIntercomHeader__container',
-        classNames?.container,
-      )}
+      className={clsx('NotifiIntercomHeader__container', classNames?.container)}
     >
       <div className={'NotifiIntercomHeader__leftContainer'}>
         <ChatIcon />
         <div
-          className={clsx(
-            'NotifiIntercomHeader__content',
-            classNames?.content,
-          )}
+          className={clsx('NotifiIntercomHeader__content', classNames?.content)}
         >
           {chatWindowHeaderContent}
         </div>
--- a/packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
@@ -24,15 +24,15 @@ export const ChatWindowHeader: React.FC<ChatWindowHeaderProps> = ({
   return (
     <div
       className={clsx(
-        'NotifiIntercomChatWindowHeader__container',
+        'NotifiIntercomHeader__container',
         classNames?.container,
       )}
     >
-      <div className={'NotifiIntercomChatWindowHeader__leftContainer'}>
+      <div className={'NotifiIntercomHeader__leftContainer'}>
         <ChatIcon />
         <div
           className={clsx(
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
```diff
@@ -24,15 +24,15 @@ export const ChatWindowHeader: React.FC<ChatWindowHeaderProps> = ({
   return (
     <div
       className={clsx(
-        'NotifiIntercomChatWindowHeader__container',
+        'NotifiIntercomHeader__container',
```
jennifer-notifi, line 27: I believe that you had the correct styling when you first created that PR and were using the old CSS classname. Can you please let me know what caused the styling to change?

packages/notifi-react-card/lib/components/intercom/SendMessageSection.tsx
```diff
@@ -31,6 +42,7 @@ export const SendMessageSection: React.FC<SendMessageSectionProps> = ({
       )}
     >
       <textarea
+        onKeyDown={handleKeypressDown}
```
jennifer-notifi, line 45: Since you are passing in the event in `handleKeypressDown`, should we update this so that it passes in the event? Something like

`onKeyDown={(e) => handleKeypressDown(e)}`

jennifer-notifi (changes requested):
Great work finding these bugs in time for UAT! Just a few notes.

packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
marukohao: yeah, I think I has changed the className in css file after the initial PR in order to use in another component, and forgot the change the className in ChatWindowHeader, should be a copy and paste mistake.

packages/notifi-react-card/lib/components/intercom/SendMessageSection.tsx
```diff
@@ -31,6 +42,7 @@ export const SendMessageSection: React.FC<SendMessageSectionProps> = ({
       )}
     >
       <textarea
+        onKeyDown={(e) => handleKeypressDown(e)}
```
juni-notifi, line 45: small, but in my opinion important, detail: we should use `keyup` rather than keydown.

packages/notifi-react-card/lib/components/intercom/ChatWindowHeader.tsx
jennifer-notifi: Got it, thank you for the clarification!

juni-notifi (commented):
Looks good, but we should use keyUp instead.

packages/notifi-react-card/lib/components/intercom/SendMessageSection.tsx
jennifer-notifi: Cool! Had to remind myself of `keyUp` vs. `keyDown`. I agree, `onKeyUp` would be better since it would happen after release

marukohao: learned😊

jennifer-notifi (approved):
LGTM! Also, looks like you have eslint and prettier set up in your IDE again!

marukohao:
> LGTM! Also, looks like you have eslint and prettier set up in your IDE again!

haha, yes, thanks Jennifer.

PR #128, opened 2022-12-01
[MVP-1697] chat window css polish

marukohao:
adjust bold dark fonts to match design: intro message and support staff ‘name’ 
Increase width of support staff message bubble
center the date component

Test:
before:
Screenshot 2022-12-01 at 10 13 27 AM
Screenshot 2022-12-01 at 9 26 39 AM

after:
Untitled

Story:
https://notifi.atlassian.net/browse/MVP-1697

### DIFF
--- a/packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
@@ -10,6 +10,7 @@ export type ChatMessageDateProps = Readonly<{
   classNames?: Readonly<{
     container: string;
     content: string;
+    body: string;
   }>;
   createdDate: string;
   isStartDate?: boolean;
@@ -29,13 +30,20 @@ export const ChatMessageDate: React.FC<ChatMessageDateProps> = ({
     >
       <div
         className={clsx(
-          'NotifiIntercomChatMessageDate__content',
-          classNames?.content,
+          'NotifiIntercomChatMessageDate__body',
+          classNames?.body,
         )}
       >
-        {isStartDate
-          ? formatConversationStartTimestamp(createdDate)
-          : formatConversationDateTimestamp(createdDate)}
+        <div
+          className={clsx(
+            'NotifiIntercomChatMessageDate__content',
+            classNames?.content,
+          )}
+        >
+          {isStartDate
+            ? formatConversationStartTimestamp(createdDate)
+            : formatConversationDateTimestamp(createdDate)}
+        </div>
       </div>
     </div>
   );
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
```diff
@@ -29,13 +30,20 @@ export const ChatMessageDate: React.FC<ChatMessageDateProps> = ({
...
       >
-        {isStartDate
-          ? formatConversationStartTimestamp(createdDate)
-          : formatConversationDateTimestamp(createdDate)}
+        <div
+          className={clsx(
+            'NotifiIntercomChatMessageDate__content',
+            classNames?.content,
+          )}
+        >
```
juni-notifi, line 42: a div with only one div as a child feels unnecessary, can you see if combining the styles for these two classnames solves your problem?

packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -635,7 +635,7 @@
 
 .NotifiIntercomHeader__content {
   font-size: 14px;
-  font-weight: 800;
+  font-weight: 650;
```
jennifer-notifi, line 638: Can you please check on the `font-weight`? 

The font in the designs in the Figma do look even more thin. 

Screen Shot 2022-12-01 at 10 50 49 AM

packages/notifi-react-card/lib/components/intercom/ChatMessageDate.tsx
marukohao: hmm, it doesn't work when combine the styles, it should work with inline-block, and width with 100%, but as the initial div container is the container with background of the date component, so I can't make width equals to 100%.

packages/notifi-react-card/lib/components/subscription/defaults.css
marukohao: the font-weight in design is 800, which is I used and looks much thick than figma, and with 600, it is thinner than in figma, so I changed to 650, not exactly the same but similar. here is the screen shot for font-weight 600:
Screenshot 2022-12-01 at 11 08 12 AM

jennifer-notifi: I think 600 does look better. Maybe send this screenshot to Sam and see what she thinks?

juni-notifi (approved):

packages/notifi-react-card/lib/components/subscription/defaults.css
marukohao: updated with 600 and also send a screen shot to Sam, will merge this in for now and if we need to change font weight, I'll create another PR. but I think it does look better too:)

PR #131, opened 2022-12-02
[MVP-1701] show support staff avatar

marukohao:
add backgroundImageUrl as createSupportConversation mutation response

Test:
Screenshot 2022-12-01 at 9 39 21 PM

Story:
https://notifi.atlassian.net/browse/MVP-1701

### DIFF
--- a/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
+++ b/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
@@ -22,6 +22,8 @@ mutation createSupportConversation {
         id
         preferredAddress
         preferredName
+        avatarData
+        avatarDataType
       }
       resolvedName
     }
--- a/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
+++ b/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
@@ -16,15 +16,16 @@ mutation createSupportConversation {
     }
     name
     createdDate
-     participants {
-      conversationParticipantType
-       profile {
-         id
-          preferredAddress
-          preferredName
-       }
-       resolvedName
-     }
+    participants {
+    conversationParticipantType
+      profile {
+        id
+        preferredAddress
+        preferredName
+      }
+      resolvedName
+    }
+    backgroundImageUrl
   }
 }
 `.trim();
--- a/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
+++ b/packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
@@ -17,7 +17,7 @@ mutation createSupportConversation {
     name
     createdDate
     participants {
-    conversationParticipantType
+      conversationParticipantType
       profile {
         id
         preferredAddress
--- a/packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
@@ -92,9 +92,11 @@ export const IntercomCard: React.FC<
           if (participant.conversationParticipantType === 'MEMBER') {
             setUserId(participant.profile.id);
           }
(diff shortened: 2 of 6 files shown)


packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -813,3 +807,7 @@
   text-align: right;
   color: #1448F3;
 }
+
+.NotifiIntercomChatMessage__avatar {
```
juni-notifi, line 811: can we also add a background-color in case the image is a broken link?

(would be good to see a screenshot of that too)

juni-notifi (approved):
Awesome!

packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
```diff
@@ -16,15 +16,16 @@ mutation createSupportConversation {
...
-          preferredAddress
-          preferredName
-       }
-       resolvedName
-     }
+    participants {
+    conversationParticipantType
+      profile {
+        id
+        preferredAddress
```
nimesh-notifi, line 23: I think it should use AvatarData and AvatarDataType here, right? It just happens that backgroundImageUrl is the same :)

packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
```diff
@@ -16,15 +16,16 @@ mutation createSupportConversation {
...
-      conversationParticipantType
-       profile {
-         id
-          preferredAddress
-          preferredName
-       }
-       resolvedName
-     }
+    participants {
+    conversationParticipantType
```
juni-notifi, line 20: indentation is wonky here too

packages/notifi-axios-adapter/lib/mutations/createSupportConversationImpl.ts
marukohao: update to use AvatarData as the url link for support staff avatar

packages/notifi-react-card/lib/components/subscription/defaults.css
marukohao: hmm, tried to apply a background and it doesn't work, add onError in  to check if the src is invalid, it will show noting instead of a broken image.
Screenshot 2022-12-01 at 10 40 23 PM

packages/notifi-react-card/lib/components/intercom/IntercomCard.tsx
```diff
@@ -91,6 +92,9 @@ export const IntercomCard: React.FC<
           if (participant.conversationParticipantType === 'MEMBER') {
             setUserId(participant.profile.id);
           }
+          if (participant.conversationParticipantType === 'SUPPORT') {
+            setSupportAvatarUrl(participant.profile.avatarData);
```
nimesh-notifi, line 96: You may want to double check that the avatardatatype is url here before setting it.

packages/notifi-react-card/lib/components/subscription/defaults.css
juni-notifi: actually this feels better

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -55,6 +55,8 @@ export type NotifiSubscriptionData = Readonly<{
   setConversationId: (conversationId: string) => void;
   userId: string;
   setUserId: (userId: string) => void;
+  setSupportAvatarUrl: (supportAvatarUrl: string) => void;
+  supportAvatarUrl: string;
```
juni-notifi, line 59: hmmmmmmm instead of pulling it from the `conversation.participant` I think both of these fields is better to pull from `conversationMessage.conversationParticipant`

For now, I'm okay with it as-is. Sorry I missed this in previous reviews.

juni-notifi (approved):

PR #134, opened 2022-12-05
update to get avatarData and avatarDataType from conversationMessage.

marukohao:
update to get avatarData and avatarDataTYpe for showing support staff avatar from conversationMessage.conversationParticipant instead of from conversation.participant

test:
Screenshot 2022-12-05 at 2 44 30 PM

### DIFF
--- a/packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
+++ b/packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
@@ -44,24 +44,21 @@ export const MessageGroup: React.FC<MessageGroupProps> = ({
             key={index}
           >
             {isIncoming ? (
-              <div>
-                <img
-                  height={34}
-                  src={
-                    participantProfile.avatarDataType === 'URL'
-                      ? participantProfile.avatarData
-                      : ''
-                  }
-                  onError={(e) => {
-                    const img = e.target as HTMLImageElement;
-                    img.style.display = 'none';
-                  }}
-                  className={clsx(
-                    'NotifiIntercomChatMessage__avatar',
-                    classNames?.avatar,
-                  )}
-                />
-              </div>
+              <img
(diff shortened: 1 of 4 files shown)


packages/notifi-react-card/lib/components/intercom/MessageGroup.tsx
```diff
@@ -34,63 +31,70 @@ export const MessageGroup: React.FC<MessageGroupProps> = ({
...
-              <div
-                className={clsx(
-                  'NotifiIntercomChatOutgoingMessage__sender',
-                  classNames?.sender,
-                )}
-              >
-                {message.conversationParticipant.resolvedName}
+              <div>
+                <img
+                  height={34}
```
juni-notifi, line 49: rather than embedding this inside a div, it would be better to use `display: block` on the img
 
also, instead of an explicit height, we should provide that in the default styles.

jennifer-notifi:
Hey Hao, just for my understanding, why should we get `avatarData` and `avatarDataType` for showing support staff avatar from` conversationMessage.conversationParticipant` instead of from `conversation.participant`?

marukohao:
> Hey Hao, just for my understanding, why should we get `avatarData` and `avatarDataType` for showing support staff avatar from `conversationMessage.conversationParticipant` instead of from `conversation.participant`?

Hi Jennifer, cause in one support chat thread, there may be more than one support staff, in conversation.participants, it will return all the participants in this thread, and we need to add logics to decide with avatarData is for which support staff. 
For data return from the conversationMessages, the conversationParticipant is the only participant in each message node, it's more easier to get the current participant's avatar.

conversation:
Screenshot 2022-12-05 at 5 58 37 PM
conversationMessages:
Screenshot 2022-12-05 at 6 02 04 PM

juni-notifi (approved):

PR #135, opened 2022-12-06
creates a context wrapper

Nathan-Notifi:
to approach creating a more globalized context where users can use hooks throughout the application, we'd need to open up the context to more than just a card. this will allow the user to create a separate context that use react card components without having to use the notifi react card

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionContextWrapper.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionContextWrapper.tsx
@@ -1,0 +1,22 @@
+import {
+  NotifiSubscriptionContextProvider,
+  useNotifiClientContext,
+} from 'notifi-react-card/lib';
+import React from 'react';
+
+export type NotifiSubscriptionContextWrapper = Readonly<{
+  cardId: string;
+}>;
+
+export const NotifiSubscriptionContextWrapper: React.FC<
+  React.PropsWithChildren<NotifiSubscriptionContextWrapper>
+> = (props: React.PropsWithChildren<NotifiSubscriptionContextWrapper>) => {
+  const { params } = useNotifiClientContext();
+  const { children } = props;
+
+  return (
+    <NotifiSubscriptionContextProvider {...params}>
+      {children}
+    </NotifiSubscriptionContextProvider>
+  );
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -36585,7 +36585,7 @@
         "@notifi-network/notifi-axios-adapter": "^0.42.1",
         "@notifi-network/notifi-axios-utils": "^0.42.1",
         "@notifi-network/notifi-core": "^0.42.1",
-        "@polkadot/util": "*",
+        "@polkadot/util": "^10.1.14",
         "@types/node": "^17.0.21",
         "@types/react": "^17.0.39",
         "axios": "^0.26.0",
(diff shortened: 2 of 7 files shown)


packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionContextWrapper.tsx
```diff
@@ -0,0 +1,28 @@
...
+  NotifiSubscriptionContextProvider,
+  useNotifiClientContext,
+} from 'notifi-react-card/lib';
+import React from 'react';
+
+import { NotifiInputLabels, NotifiInputSeparators } from '.';
+
+export type NotifiSubscriptionContextWrapper = Readonly<{
+  inputLabels?: NotifiInputLabels;
+  darkMode?: boolean;
```
Nathan-Notifi, line 11: i do think these props will be helpful to use in other components as we build but not sure how it will be for our current integration

packages/notifi-react-example/package.json
```diff
@@ -48,5 +48,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
+    "process": "^0.11.10"
```
jennifer-notifi, line 53: Can you please advise why we need to add `process `as a dev dependency? I believe the webpack file does handle the polyfill error relating to `process`.

packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionContextWrapper.tsx
jennifer-notifi: Can you please elaborate on this more? I was curious as to why we may still be typing the wrapper the same way that we would type the SubscriptionCard. 

Is the idea that integrators will eventually pass in their own type of card?

packages/notifi-react-example/package.json
Nathan-Notifi: ah, this was something I noticed while trying to get the card running

packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
```diff
@@ -67,6 +68,7 @@ export const NotifiCard: React.FC = () => {
           inputSeparators={intercomInputSeparators}
           cardId="1045f61752b148eabab0403c08cd60b2"
         />
+        <NotifiSubscriptionContextWrapper cardId="1045f61752b148eabab0403c08cd60b2"></NotifiSubscriptionContextWrapper>
```
jennifer-notifi, line 71: Can you please update this to 
`<NotifiSubscriptionContextWrapper cardId="1045f61752b148eabab0403c08cd60b2" />`

We probably shouldn't commit testing updates in notifi-react-example but I think it is also okay for now.

Nathan-Notifi: i'll just back out the notfii example changes

jennifer-notifi:
A better question to passing `darkMode` is how would the integrator utilize our CSS and be able to apply `darkMode` to their children?

Nathan-Notifi:
I think that could be similar in applying it to other situations.
you could use a hook to determine if the boolean is on and then wrap it.

      className={clsx(
        darkMode ? 'notifi__dark' : 'notifi__light',
        'NotifiSubscriptionCard__container',
        classNames?.container,
      )}

this is at the top level. they could use our css.

jennifer-notifi:
> I think that could be similar in applying it to other situations. you could use a hook to determine if the boolean is on and then wrap it.
> 
> ```
>   className={clsx(
>     darkMode ? 'notifi__dark' : 'notifi__light',
>     'NotifiSubscriptionCard__container',
>     classNames?.container,
>   )}
> ```
> 
> this is at the top level. they could use our css.

Yes, I think that having some parameters at the top level may be useful - I think we should evaluate as we have more users using the wrapper.

jennifer-notifi (approved):
LGTM

PR #137, opened 2022-12-07
Sets up twitter and discord inputs, adds more flexibility to separate cards

Nathan-Notifi:
sets up the cards to support twitter and discord
- added some additional fields so that the cards can be used separately in the future. 
- discord and twitter might be undefined so added them to the config so that it doesn't break anything
- adds ability to hide start icons


packages/notifi-react-card/lib/components/subscription/defaults.css
```diff
@@ -6,7 +6,7 @@
   --notifi-secondary-font-color: rgb(93, 93, 93);
   --notifi-input-background: rgba(255, 255, 255, 0.8);
   --notifi-intercom-input-background: rgb(245, 246, 251, 1);
-  --notifi-placeholder-color: rgba(224, 224, 224, 0.5);
+  --notifi-placeholder-color: rgb(33 33 33 / 50%);
```
juni-notifi, line 9: woah, what's this syntax?

packages/notifi-react-card/lib/components/subscription/subscription-card-views/InputFields.tsx
```diff
@@ -0,0 +1,186 @@
...
+            className={clsx(
+              'NotifiInputSeparator__content',
+              inputSeparators.telegramSeparator.classNames?.content,
+            )}
+          >
+            {inputSeparators?.telegramSeparator?.content}
+          </div>
+        </div>
+      ) : null}
+      {data.contactInfo.discord?.active ? (
```
juni-notifi, line 134: have you added the `discord` and `twitter` fields to the contact info type? I can't seem to find it in this diff

Nathan-Notifi: image
this? @juni-notifi

packages/notifi-react-card/lib/components/subscription/defaults.css
Nathan-Notifi: ohh it's another way for opacity. the current light one was a little too hard to read

packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
```diff
@@ -92,7 +94,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
           classNames?.container,
         )}
       >
-        <EmailIcon className={'NotifiInput__icon'} />
+        {hideStartIcon ? null : <EmailIcon className={'NotifiInput__icon'} />}
```
juni-notifi, line 97: now that i think about it, i think this should be done by overriding CSS --

the user can provide a stylesheet with
```
.NotifiInput__icon {
  display: none;
}
```

I think this is more flexible than a prop, since it's harder to provide a specific prop to the inputs if you're using the Subscription card. wdyt?

packages/notifi-react-card/lib/components/subscription/subscription-card-views/PreviewCard.tsx
```diff
@@ -28,7 +29,9 @@ export type PreviewCardProps = Readonly<{
 export const PreviewCard: React.FC<PreviewCardProps> = ({
   classNames,
   data,
+  hideBackClick,
   inputDisabled,
+  hideEditButton,
```
juni-notifi, line 34: similar to the input icon, i think we can achieve this with CSS overrides rather than props

packages/notifi-react-card/lib/components/subscription/subscription-card-views/preview-panel/UserInfoPanel.tsx
```diff
@@ -148,15 +154,51 @@ export const UserInfoPanel: React.FC<UserInfoPanelProps> = ({
...
-      <button
-        className={clsx(
-          'NotifiPreviewCard__editButton',
-          classNames?.EditButton,
-        )}
-        onClick={handleEditClick}
-      >
-        Edit
-      </button>
+      {contactInfo?.discord?.active ? (
```
juni-notifi, line 157: contactInfo itself isn't nullable is it?

let's update to `contactInfo.discord?.active`

packages/notifi-react-card/lib/components/subscription/subscription-card-views/EditCardView.tsx
```diff
@@ -45,86 +42,25 @@ export const EditCardView: React.FC<EditCardViewProps> = ({
...
+  hideStartIcons,
 }) => {
   return (
     <div
       className={clsx('NotifiInputContainer', classNames?.NotifiInputContainer)}
     >
-      <AlertListPreview eventTypes={data.eventTypes} />
+      {hideAlertListPreview ? null : (
+        <AlertListPreview eventTypes={data.eventTypes} />
+      )}
```
juni-notifi, line 54: ditto -- i think we should try to achieve this with CSS overrides rather than props.

juni-notifi (commented):
I think we should avoid having props for some of these.

packages/notifi-react-card/lib/components/subscription/subscription-card-views/preview-panel/UserInfoPanel.tsx
Nathan-Notifi: addressed

packages/notifi-react-card/lib/components/subscription/subscription-card-views/PreviewCard.tsx
Nathan-Notifi: let me try that

packages/notifi-react-example/src/NotifiCard/NotifiCard.css
```diff
@@ -2,3 +2,23 @@
...
+  display: none
+}
+
+.NotifiPreviewCard__backArrowContainer{
+  display: none;
+}
+
+/* .NotifiPreviewCard__editButton{
+  display: none;
+} */
```
juni-notifi, line 24: oh we probably don't want this committed to react example

juni-notifi (approved):

packages/notifi-react-example/src/NotifiCard/NotifiCard.css
Nathan-Notifi: removed!

PR #140, opened 2022-12-10
Set state to 'edit' if data is missing

juni-notifi:
For some reason the card was always going to the alert history view.


Nathan-Notifi (approved):

jennifer-notifi (approved):
Thank you for the update, Juni. Also, thank you to Nathan for the initial review.

I confirmed using the notifi-react-example that:

1) with an existing signed-up wallet, show Alert History
2) with a new wallet, it goes to the edit page

Long-term, would it make more sense to show alert history if the user has signed up AND has alerts?

PR #141, opened 2022-12-13
Input Form Context Update

Nathan-Notifi:
Components relying on the subscription context can display the wrong data if a user updates the form and somehow can navigate to another page.

in a future update, we should decouple how the user views data and how the user updates data so that data is presented more accurately.

this was by creating a new form context that can be updated via the inputs and then accessed on subscription.

### DIFF
--- a/packages/notifi-react-card/lib/components/NotifiSmsInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiSmsInput.tsx
@@ -57,6 +57,24 @@ export const NotifiSmsInput: React.FC<NotifiSmsInputProps> = ({
     dialCode: '+1',
   });
 
+  const handleDialCodeChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>) => {
+      setPhoneValues({ ...phoneValues, dialCode: event.target.value });
+      setPhoneNumber(event.target.value + phoneValues.baseNumber);
+    },
+    [phoneValues, setPhoneNumber, setPhoneValues],
+  );
+
+  const handleBaseNumberChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>) => {
+      const onlyNumberInput = event.target.value.replace(/[^\d]/g, '');
+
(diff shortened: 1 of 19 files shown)


packages/notifi-react-card/lib/components/NotifiSmsInput.tsx
```diff
@@ -76,52 +84,59 @@ export const NotifiSmsInput: React.FC<NotifiSmsInputProps> = ({
   }, [phoneNumber, splitPhoneValues]);
 
   useEffect(() => {
-    if (phoneValues.baseNumber === '') {
+    if (phoneValues.baseNumber === '' && phoneNumber !== '') {
+      console.log('phone number', phoneNumber);
```
juni-notifi, line 88: remove?

packages/notifi-react-card/lib/components/NotifiSmsInput.tsx
```diff
@@ -76,52 +84,59 @@ export const NotifiSmsInput: React.FC<NotifiSmsInputProps> = ({
...
-        setPhoneNumber(event.target.value + phoneValues.baseNumber);
-      } else {
-        setPhoneNumber(phoneValues.dialCode + event.target.value);
-      }
-    };
+          setPhoneValues({ ...phoneValues, [field]: onlyNumberInput });
+        } else {
+          setPhoneValues({ ...phoneValues, [field]: event.target.value });
+        }
+        if (field === 'dialCode') {
```
juni-notifi, line 132: the fact that we have `if (field)` here makes me think that we should just have two different functions, handleDialCodeChange and handleBaseNumberChange, rather than parameterizing the field (which results in a different function instance actually used for the event handler on every render)

juni-notifi (approved):
kind of weird that we depend on intercomCardView in the telegram view, let's clean that up later

packages/notifi-react-card/lib/components/NotifiSmsInput.tsx
Nathan-Notifi: genius

juni-notifi (approved):

PR #144, opened 2022-12-16
small updates for positioninig and button colors

Nathan-Notifi:
we have position absolute on a child and the parent didn't have any position prop so it was using the viewport width the edit button color also didn't have enough contrast so i updated it to match the subscribe button

- updated settings icons colors through through property 
- added a max height to alert details container, it was spilling over without it.

Before:
image

After:
image

Header:
Before:
image

image

Settings Icons Colors:
image

image

max-height before:
image

max-height after:
image

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -155,6 +155,7 @@ input:-webkit-autofill {
   justify-content: center;
   align-items: center;
   padding-bottom: 10px;
+  position: relative;
 }
 
 .NotifiAlertHistory__notificationDate {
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -482,7 +483,7 @@ input:-webkit-autofill {
   font-size: 16px;
   line-height: 20px;
   cursor: pointer;
-  color: var(--notifi-font-color);
+  color: #ffffff;
```
jennifer-notifi, line 486: Should we have a --notifi-button-font-color that defaults to white then?

Nathan-Notifi: sure

juni-notifi (approved):

PR #149, opened 2022-12-22
[MVP-1810] add liquidation alert selection to SDK react card - UI

marukohao:
Add Liquidation LTV Alert Selection to SDK React Card
Toggle: If toggle is turned ON, then the notification buttons drop down. The user can select or enter per 
Button 1: XX%
Button 2: YY%
Button 3: Custom entry of a percent ranging from 1%-99%

## Test:
Untitled1

## Story: 
https://notifi.atlassian.net/jira/software/c/projects/MVP/boards/1?modal=detail&selectedIssue=MVP-1810&sprint=63

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 
 import { useNotifiSubscriptionContext } from '../../context';
 import { CheckRatio, HealthCheckEventTypeItem } from '../../hooks';
@@ -7,6 +7,13 @@ import { DeepPartialReadonly } from '../../utils';
 import type { NotifiToggleProps } from './NotifiToggle';
 import { NotifiToggle } from './NotifiToggle';
 import { NotifiTooltip, NotifiTooltipProps } from './NotifiTooltip';
+
+const getParsedInputNumber = (input: string): number | null => {
+  if (input.indexOf('%') === input.length - 1) {
+    return parseFloat(input.slice(0, -1)) ?? null;
+  }
+  return null;
+};
 
 export type EventTypeHealthCheckRowProps = Readonly<{
   classNames?: DeepPartialReadonly<{
@@ -30,35 +37,30 @@ export const EventTypeHealthCheckRow: React.FC<
   // const { instantSubscribe } = useNotifiSubscribe({
   //   targetGroupName: 'Default',
   // });
-  const [enabled, setEnabled] = useState(false);
-  const [ratios, setRatios] = useState<CheckRatio[]>([]);
+  // const [, setEnabled] = useState(false);
   const [selectedIndex, setSelectedIndex] = useState<number | null>(null);
   const [selectedRatio, setSelectedRatio] = useState<number | null>(null);
   const [customValue, setCustomValue] = useState<string>('');
-  const [cutomButtonPlaceholder, setCustomButtonPlaceholder] =
-    useState<string>('Custom');
   const [errorMessage, setErrorMessage] = useState<string>('');
   const [isShowSelectButton, setIsShowSelectButton] = useState<boolean>(false);
 
   const alertName = useMemo<string>(() => config.name, [config]);
   const tooltipContent = config.tooltipContent;
 
-  useEffect(() => {
-    if (loading) {
-      return;
-    }
-    const hasAlert = alerts[alertName] !== undefined;
-    setEnabled(hasAlert);
-  }, [alertName, alerts, loading]);
+  const isValueType = config && config.checkRatios.type === 'value';
+  let ratios: CheckRatio[] = [];
+  if (isValueType) {
+    ratios = config.checkRatios.value;
+  }
+
+  const enabled = !loading && !!alerts[alertName];
 
   useEffect(() => {
-    if (config && config.checkRatios.type === 'value') {
-      const ratios = config.checkRatios.value;
-      setRatios(ratios);
+    if (isValueType) {
       setSelectedRatio(ratios[-1]?.ratio);
       setSelectedIndex(ratios.length - 1);
     }
-  }, [config, setRatios]);
+  }, [isValueType]);
 
   const handleToggleNewSubscription = () => {
     if (loading) {
@@ -91,7 +93,7 @@ export const EventTypeHealthCheckRow: React.FC<
       // instantSubscribe({
       //   alertConfiguration: healthThresholdConfiguration({
       //     alertFrequency: 'ALWAYS',
-      //     percentage: ratioNumber,
+      //     percentage: selectedRatio,
       //   }),
       //   alertName: alertName,
       // });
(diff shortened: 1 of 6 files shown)


packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
```diff
@@ -39,11 +38,11 @@ export const NotifiCard: React.FC = () => {
     },
   };
 
-  const intercomInputSeparators: NotifiInputSeparators = {
-    emailSeparator: {
-      content: 'OR',
-    },
-  };
+  // const intercomInputSeparators: NotifiInputSeparators = {
```
jennifer-notifi, line 41: We should probably not commit any tests to `notifi-react-example`. Hmm, we should probably add `notifi-react-example` to `.gitignore` at some point.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,205 @@
...
+    if (!isShowSelectButton) {
+      setIsShowSelectButton(true);
+      setEnabled(true);
+    } else {
+      if (
+        customValue.indexOf('%') === customValue.length - 1 &&
+        parseFloat(customValue.slice(0, -1)) >= 1 &&
+        parseFloat(customValue.slice(0, -1)) <= 99
+      ) {
+        console.log(selectedRatio);
```
jennifer-notifi, line 78: Can you please remove this `console.log`?

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,205 @@
...
+                    setErrorMessage('');
+                  }}
+                >
+                  {percentage}
+                </div>
+              );
+            })}
+            <input
+              onClick={() => {
+                //assume we should only allow no more than two specific values and one custom value here
```
jennifer-notifi, line 168: Agreed here. I also was thinking if a tenant would allow a user to add a custom value and if that option should be configured in the config tool.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const handleNewSubscription = () => {
+    if (loading) {
+      return;
+    }
+    if (!isShowSelectButton) {
+      setIsShowSelectButton(true);
+      setEnabled(true);
+    } else {
+      if (
+        customValue.indexOf('%') === customValue.length - 1 &&
```
jennifer-notifi, line 74: Should the custom value be between the values that the tenant inputted through the config? Or any values between 0 and 1?

jennifer-notifi (commented):
Here are initial comments, I will finish reviewing thoroughly after lunch!

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
marukohao: reference from the ticket: Button 3: Custom entry of a percent ranging from 1%-99%

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
marukohao: I'll confirm with Sam and Juni and will address this in next PR(connect API)

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
jennifer-notifi: Got it, thank you!

packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
```diff
@@ -24,6 +24,25 @@ export type BroadcastEventTypeItem = Readonly<{
   tooltipContent?: string;
 }>;
 
+export type HealthCheckEventTypeItem = Readonly<{
```
jennifer-notifi, line 27: Nothing to change: We need a type library that we can use between the config tool and SDK. No need to build it out now but something that would really help if we ever have time to build it out in the future (key word is 'ever' which pretty means 'never', haha).

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,209 @@
...
+    setEnabled(hasAlert);
+  }, [alertName, alerts, loading]);
+
+  useEffect(() => {
+    if (config && config.checkRatios.type === 'value') {
+      setRatios(config.checkRatios.value);
+    }
+  }, [config, setRatios]);
+
+  const handleNewSubscription = () => {
```
jennifer-notifi, line 65: Ah, I believe you should handle if a subscriber turns on the toggle and then wants to turn off the toggle without entering a custom.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,209 @@
...
+    if (loading) {
+      return;
+    }
+    if (!isShowSelectButton) {
+      setIsShowSelectButton(true);
+      setEnabled(true);
+    } else {
+      const selectedRatioNumber = parseFloat(customValue.slice(0, -1));
+      if (
```
jennifer-notifi, line 74: Similar here, I think we should handle the check if the user clicks out of the input, not when they uncheck the input. 

I wonder if it would be better to also update NotifiToggle to allow passing a more general function as well. I ran into several issues toggling back and forth.

Currently, NotifiToggle accepts setChecked which is supposed to only update if it is checked or not:

```
onChange={(e) => {
          setChecked(e.target.checked);
        }}
```

I am not sure what the customer impact would be though!

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,209 @@
...
+        selectedRatio !== '' ||
+        (customValue.indexOf('%') === customValue.length - 1 &&
+          selectedRatioNumber >= 1 &&
```
jennifer-notifi, line 77: Very nit (you can do it in the next PR): Will the user run into issues if they add a % sign at the end of the number?

jennifer-notifi (changes requested):

jennifer-notifi:
Also, I assume that any ratio that I choose is not persisting when I go back to the card because we have not implemented the API yet. Please confirm if that is the current intended behavior.

jennifer-notifi:
You're almost there! I would suggest taking a second look at the following behaviors:

1. If a user toggles on the liquidation alert and then toggles off, make sure that they can toggle off
2. If a user toggles on the liquidation alert, types in a custom ratio or chooses any ratio, they can then toggle off if they want to
3. If a user toggles on the liquidation alert, types in a custom ratio with a percentage or any other special characters, we either handle it or give them an error message

marukohao:
> You're almost there! I would suggest taking a second look at the following behaviors:
> 
> 1. If a user toggles on the liquidation alert and then toggles off, make sure that they can toggle off
> 2. If a user toggles on the liquidation alert, types in a custom ratio or chooses any ratio, they can then toggle off if they want to
> 3. If a user toggles on the liquidation alert, types in a custom ratio with a percentage or any other special characters, we either handle it or give them an error message

Hi Jenn, confirmed with Sam, and here is the flow for create alert:
if they toggle it on, we create the alert and auto select the center button 
2. user toggles on
3. the alert is now on and set to 70%, the subtext and % options are displayed in the on state 
4. user can adjust the % or toggle back off

I think we can assume that if the user click on any of the ratio select button, the alert will be update with the selected value, and for custom button, I am current create the alert when the user hit enter or focus on anything outside of the input button(should double check with Sam on this, and will have addressed in the next PR).

jennifer-notifi (commented):
Thank you for checking on the flow and the clean-up - the code looks a lot more clean!

I confirmed that when I toggle on for the first time, the default selection is the second, pre-defined LTV or the middle button. 

I think that there is still some polish that we can pursue.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,248 @@
...
+      setSelectedRatio(ratios[-1]?.ratio);
+      setSelectedIndex(ratios.length - 1);
+    }
+  }, [config, setRatios]);
+
+  const handleToggleNewSubscription = () => {
+    if (loading) {
+      return;
+    }
+    setIsShowSelectButton(!isShowSelectButton);
```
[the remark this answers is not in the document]
marukohao: yeah, the boolean value of enabled is based on if this alert exist or not, we can't have the toggle work as expect until connect API for subscribe.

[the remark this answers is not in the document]
jennifer-notifi: Got it, thank you for the clarification. Will mark this as resolved.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -0,0 +1,248 @@
...
+    }
+  };
+
+  const handleCustomRatioButtonNewSubscription = () => {
+    if (loading) {
+      return;
+    }
+    if (selectedRatio) {
+      if (
+        customValue.indexOf('%') === customValue.length - 1 &&
```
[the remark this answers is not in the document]
marukohao: good catch, this should be fixed

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
marukohao: asked Juni, we should limited the check ratios no more than two from the admin config when tenant set up. I'll open a PR to address it in admin tool

jennifer-notifi: Thank you for checking!

jennifer-notifi (approved):
Thank you for being open to the suggestions and the continued code polish! 

Tested and confirmed the following cases:
1. When a user subscribes to the liquidation alert for the first time, they default to the second LTV or middle button.
2. I can enter in any number between 1% to 99% in custom. I cannot enter in any special characters, including % or % with other characters.

PR #151, opened 2022-12-27
[MVP-1828] connect API for creating liquidation alert

marukohao:
user need to create the liquidation alert on react subscription card
hook up with create/update alert API when user toggles the button

## Testing:
Untitled1

## Story:
https://notifi.atlassian.net/jira/software/c/projects/MVP/boards/1?modal=detail&selectedIssue=MVP-1828

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
@@ -73,17 +73,22 @@ export const EventTypeHealthCheckRow: React.FC<
     const alert = alerts[alertName];
     const checkRatios = ratios.map((ratio) => ratio.ratio);
     if (alert) {
-      const alertRatioValue = JSON.parse(alert.filterOptions ?? '').threshold;
+      let alertRatioValue: number | null = null;
+      if (alert.filterOptions) {
+        alertRatioValue = JSON.parse(alert.filterOptions).threshold;
+      }
       setEnabled(true);
-      ratios.forEach((ratio, index) => {
-        if (ratio.ratio === alertRatioValue && customValue === '') {
-          setSelectedIndex(index);
+      if (alertRatioValue) {
+        ratios.forEach((ratio, index) => {
+          if (ratio.ratio === alertRatioValue && customValue === '') {
+            setSelectedIndex(index);
+          }
+        });
+        setInitialRatio(alertRatioValue);
+        if (!checkRatios.includes(alertRatioValue) && customValue === '') {
+          setSelectedIndex(3);
+          setCustomValue(alertRatioValue * 100 + '%');
         }
-      });
-      setInitialRatio(alertRatioValue);
-      if (!checkRatios.includes(alertRatioValue) && customValue === '') {
-        setSelectedIndex(3);
-        setCustomValue(alertRatioValue * 100 + '%');
       }
     } else {
       setEnabled(false);
@@ -179,6 +184,7 @@ export const EventTypeHealthCheckRow: React.FC<
     if (event.key === 'Enter') {
       if (customInputRef.current) {
         customInputRef.current.blur();
+        event.preventDefault();
       }
     }
   };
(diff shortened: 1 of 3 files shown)


Nathan-Notifi:
nice demo! does the number stay when refreshing the information?

marukohao:
> nice demo! does the number stay when refreshing the information?

the custom ratio will be stay if the user has saved it(clicking on outside of the input box), otherwise user has to re-enter the number.

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -53,52 +59,83 @@ export const EventTypeHealthCheckRow: React.FC<
...
-    //   });
-    // } else {
-    //   instantSubscribe({
-    //     alertConfiguration: null,
-    //     alertName: alertName,
-    //   });
-    // }
-  };
+    setErrorMessage('');
+    if (!enabled && initialRatio) {
```
juni-notifi, line 101: i think we should do an explicit check on `initialRatio !== null` because a user might want to set an alert at 0% health

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -108,19 +145,21 @@ export const EventTypeHealthCheckRow: React.FC<
...
-      //   alertConfiguration: healthThresholdConfiguration({
-      //     alertFrequency: 'ALWAYS',
-      //     percentage: ratioNumber,
-      //   }),
-      //   alertName: alertName,
-      // });
+    if (ratioNumber && ratioNumber >= 1 && ratioNumber <= 99 && customValue) {
+      instantSubscribe({
+        alertConfiguration: healthThresholdConfiguration({
+          alertFrequency: 'DAILY',
```
juni-notifi, line 154: now that i think about it, it would be nice if we can get the alertFrequency from the card config. Let's do that as a followup

juni-notifi (commented):
Can you also make it so that the user can press "enter" to save the value? Thanks!

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -53,52 +59,83 @@ export const EventTypeHealthCheckRow: React.FC<
...
+        alertName: alertName,
+      })
+        .then(() => {
+          setSelectedIndex(index);
+          setCustomValue('');
+        })
+        .catch(() => setErrorMessage(UNABLE_TO_SUBSCRIBE));
     } else {
-      setErrorMessage('Please enter a valid number');
+      setErrorMessage(UNVALID_NUMBER);
```
Nathan-Notifi, line 138: I think this is `invalid`

marukohao: good catch

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -53,74 +66,120 @@ export const EventTypeHealthCheckRow: React.FC<
...
   useEffect(() => {
-    if (isValueType) {
-      setSelectedRatio(ratios[-1]?.ratio);
+    if (loading) {
+      return;
+    }
+    const alert = alerts[alertName];
+    const checkRatios = ratios.map((ratio) => ratio.ratio);
+    if (alert) {
+      const alertRatioValue = JSON.parse(alert.filterOptions ?? '').threshold;
```
juni-notifi, line 76: JSON.parse will throw an error if you use `''` I think.

Let's add some safety around parsing the filterOptions

packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -53,74 +66,120 @@ export const EventTypeHealthCheckRow: React.FC<
...
+        setErrorMessage(INVALID_NUMBER);
+      }
+    }
+  };
+
+  const handleKeypressUp = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === 'Enter') {
+      if (customInputRef.current) {
+        customInputRef.current.blur();
+      }
```
juni-notifi, line 182: remember to call `preventDefault` when adding custom handlers

juni-notifi (approved):

PR #155, opened 2023-01-03
[MVP-1855] fix intercom issue with padding and empty strings

marukohao:
disabled send button when the user input string are all spaces
increase left and right padding in the message bubble
update to get email, telegram and sms for intercom from useNotifiForm
fixed the 'light mode' disabled button color

before:
Screenshot 2023-01-01 at 4 21 03 PM
Screenshot 2022-12-30 at 2 39 47 PM

after:
Screenshot 2023-01-01 at 4 20 12 PM
Screenshot 2023-01-03 at 9 22 51 AM

test:
Untitled1

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -16,11 +16,11 @@
   --notifi-input-background: #f5f6fb;
   --notifi-input-border: #e8ebf5;
   --notifi-button-color: #1448f3;
-  --notifi-button-text-color: #fff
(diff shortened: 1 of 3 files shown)


Nathan-Notifi:
can we add a little more?

image

marukohao:
> can we add a little more?
> 
> image

Cause we don't allow user to send space only messages, so I don't think we will have the empty message bubble.

Nathan-Notifi:
hm, okay. i think that makes sense to me.

Nathan-Notifi (approved):

PR #156, opened 2023-01-03
Placeholder updates

Nathan-Notifi:
We didn't have a a place to update the copy for text fields
Updated some of hte props to allow a user to pass in text.

added an additional field for `InputsFields` to include placeholder text.


marukohao (approved):
LGTM

PR #159, opened 2023-01-06
MVP-1702 country code dropdown improvement

seanndev:
## Problem
[MVP-1702](https://notifi.atlassian.net/browse/MVP-1702)

- Country code dropdown improvement
## Testing
image

[MVP-1702]: https://notifi.atlassian.net/browse/MVP-1702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -413,6 +413,108 @@ input:-webkit-autofill {
   outline: none;
   box-shadow: 0px 0px 0px 1px var(--notifi-focus-color);
 }
+.NotifiSmsInput__dropdownSelected {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  height: 100%;
+  color: var(--notifi-font-color);
+  background-color: var(--notifi-input-background);
+  margin-top: 0;
+  margin-bottom: 0 !important;
+  cursor: pointer;
+}
+
+.NotifiSmsInput__dropdownSelectValue {
+  margin-top: 0;
+  margin-bottom: 0 !important;
+  padding: 10px 25px 10px 0;
+  height: auto;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  overflow: hidden;
+}
+
+.NotifiSmsInput__dropdownInput {
+  bottom: 0px;
+  left: 0px;
+  position: absolute;
+  opacity: 0;
+  pointer-events: none;
+  width: 100%;
+  margin-bottom: 0 !important;
+}
+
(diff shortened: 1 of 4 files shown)


juni-notifi:
why doesn't the emoji render as a flag? is it a font issue?

Nathan-Notifi:
@juni-notifi  i think it might habe to do with the browser.
image
this is using chrome.

@seanndev can we center the text? otherwise it looks pretty great!

seanndev:
Hi @Nathan-Notifi, I have updated.
image

packages/notifi-react-card/lib/constants/countryData.tsx
```diff
@@ -0,0 +1,137 @@
...
+  name: string;
+  dialCode: string;
+  flag: string;
+};
+
+type CountryMap = {
+  [countryCode: string]: CountryMetadata;
+};
+
+export const countryMap: CountryMap = {
```
juni-notifi, line 11: I think `[countryCode: string]` is a weak constraint -- it allows me to invoke invalid keys, like `countryMap['hello world']`.

Consider exporting `countryMap = { ... } as const`, `type CountryMap = typeof countryMap`, and `type CountryMetadata = CountryMap[keyof CountryMap]` instead.

juni-notifi (commented):
I think the browser affecting the flag rendering adds support to my thesis that it's typeface-dependent.

If that's the case I'm cool with leaving it for now (until we can figure out a reliable font-family to use)

@seanndev  I've left a few nit-picks. Please let me know if you run into any issues implementing my recommendations.

seanndev:
Hi @juni-notifi, I have updated all.

packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -413,6 +413,111 @@ input:-webkit-autofill {
   outline: none;
   box-shadow: 0px 0px 0px 1px var(--notifi-focus-color);
 }
+.NotifiSmsInput__dropdownSelected {
+  display: flex;
+  align-items: center;
+  width: 100%;
```
Nathan-Notifi, line 419: instead of width: 100%; and height: 100%,
can we add `display: flex` to the parent property and make this flex: 1? so it can take up the container?

seanndev:
Hi @Nathan-Notifi, I have updated

seanndev:
Hi @Nathan-Notifi, i have updated

Nathan-Notifi (approved):

PR #160, opened 2023-01-09
use alertFrequency from config data to create health check alert

marukohao:
Currently, we allow tenant to select alert frequency for health check in admin tool
update in SDK subscription card to get alertFrequency for health check from config data


packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
```diff
@@ -1,4 +1,5 @@
 // TODO: Import from library rather than copy / paste
+import { AlertFrequency } from 'notifi-frontend-client/dist';
```
juni-notifi, line 2: this is an incorrect import~ import from `@notifi-network/notifi-core` instead

btw, any time you see a `/dist` import, that's not right. It's due to Visual Studio using relative imports.

marukohao: ah, good catch! updated

juni-notifi (approved):

PR #161, opened 2023-01-10
add thredholdDirection in FilterOptions for creating health check alert

marukohao:
we need to add thredholdDirection('below' or 'above') in FilterOptions when creating the health check alert
also add logics for showing 'over' or 'under' threshold in alert history for health check

## test:
Untitled1
Screenshot 2023-01-10 at 3 21 25 PM
Screenshot 2023-01-10 at 3 21 16 PM


packages/notifi-react-card/lib/components/subscription/EventTypeHealthCheckRow.tsx
```diff
@@ -62,8 +63,10 @@ export const EventTypeHealthCheckRow: React.FC<
 
   const isValueType = config && config.checkRatios.type === 'value';
   let ratios: CheckRatio[] = [];
+  let thresholdDirection: ThresholdDirection = 'below';
   if (isValueType) {
     ratios = config.checkRatios.value;
+    thresholdDirection = ratios[0].type;
```
juni-notifi, line 69: bigger than the scope of this ticket, but we should probably put "direction" on the HealthCheckEventType, instead of inside a CheckRatio

juni-notifi (approved):

PR #162, opened 2023-01-11
Prepopulate form

Nathan-Notifi:
When a user does not have chat alerts set up,
the form should be prepopulated with the user information if they currently have the SDK setup on their application.

how to test:

create a new wallet
sign up for sdk alerts on react card
refresh on intercom card to see if user info is pre-populated


marukohao (approved):
tested and looks good!

PR #163, opened 2023-01-12
Implement tradingPair renderer

juni-notifi:
For now, add a tradingPair renderer which subscribes to Direct Tenant Messages.

This is the first instance where we have one "EventItem" that resutls in many Alerts. We need to be able to associate it to the one EventItem, so for now encode the information within alert.name.

Screen Shot 2023-01-11 at 9 41 55 PM
Screen Shot 2023-01-11 at 9 42 13 PM
Screen Shot 2023-01-11 at 9 41 42 PM

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -1183,15 +1183,35 @@ input:-webkit-autofill {
   padding: 10px 0;
 }
 
-.TradingPairAlertRow__name {
+.TradingPairAlertRow__textContainer {
+  display: flex;
   flex-grow: 1;
+  flex-direction: column;
+  align-items: stretch;
+  margin-right: 10px;
   margin-bottom: 0 !important;
+}
+
+.TradingPairAlertRow__name {
+  display: block;
+  font-size: 16px;
+  line-height: 20px;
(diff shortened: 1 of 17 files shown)


packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -1079,3 +1078,180 @@ input:-webkit-autofill {
...
 }
+
+.EventTypeTradingPairsRow__container {
+  display: flex;
+  align-items: stretch;
+  flex-direction: column;
+}
+
+.EventTypeTradingPairsRow__container > * {
+  margin-bottom: 0 !important;
```
Nathan-Notifi, line 1089: i notice we do this a lot. i wonder if it may have been better to address the :not(:last-child) instead haha

marukohao (approved):
LGTM, will we also need to add 'tradingPair' event to config tool?

packages/notifi-react-card/lib/components/defaults.css
juni-notifi: haha yeah probably, we should clean it up

juni-notifi:
> LGTM, will we also need to add 'tradingPair' event to config tool?

I think we need some more definition before we can expose it in the config tool.

PR #166, opened 2023-01-14
Implement createTenantBalanceChangeAlert

juni-notifi:
Key on the `name` parameter to create an alert with a BALANCE filter using the webhook parameters that are passed in. The method tries its best to return (optimistically) the existing entities when appropriate.

Remaining work:
- Update this method not to return Sources from SourceGroup (probably)
- Expose Sources connection
- Expose addSourceToSourceGroup / removeSourceFromSourceGroup

### DIFF
--- a/packages/notifi-node/lib/client/NotifiClient.ts
+++ b/packages/notifi-node/lib/client/NotifiClient.ts
@@ -1,7 +1,4 @@
-import type {
-  Types as Gql,
-  NotifiService,
-} from '@notifi-network/notifi-graphql';
+import { Types as Gql, NotifiService } from '@notifi-network/notifi-graphql';
 
 import type {
   Authorization,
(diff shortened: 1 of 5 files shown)


packages/notifi-node/lib/client/NotifiClient.ts
```diff
@@ -226,6 +223,214 @@ class NotifiClient {
...
+
+  createOrUpdateWebhook: (
+    params: Gql.CreateWebhookTargetMutationVariables,
+  ) => Promise<Gql.WebhookTargetFragmentFragment> = async (params) => {
+    const getResult = await this.service.getWebhookTargets({});
+    const existing = getResult.webhookTarget?.find((w) => {
+      return w.url === params.url && w.format === params.format;
+    });
+
+    if (existing === undefined) {
```
nimesh-notifi, line 373: I wonder if we should just move to a model where the create functions always return an existing object on attempt to create a dupe. Just to simplify and reduce rtt

nimesh-notifi (approved):

packages/notifi-node/lib/client/NotifiClient.ts
juni-notifi: i think that would be nice

one consideration: the client wouldn't have the `id`, so we would need to move that logic to using `name` as the identifier somehow (we dedupe on `name` inside the SDK).

PR #178, opened 2023-02-16
Add Multi-Wallet support

juni-notifi:
In e.g. Wallet Extension settings, we need to be able to support multiple connected wallets (and multiple private keys). Expose a new Multi-Wallet paradigm where we expose the list of "owned" wallets to the user and ask them to individually connect them.


packages/notifi-core/lib/operations/ConnectWallet.ts
```diff
@@ -0,0 +1,19 @@
+import { ConnectedWallet, Operation, WalletBlockchain } from '../models';
+
+export type ConnectWalletInput = Readonly<{
+  walletPublicKey: string;
+  timestamp: number;
+  signature: string;
+  walletBlockchain: WalletBlockchain;
+  accountId?: string;
```
Nathan-Notifi, line 8: what happens if the account id is undefined?

packages/notifi-react-card/lib/assets/PenIcon.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+export type Props = Readonly<{
+  className?: string;
+}>;
+
+export const PenIcon: React.FC<Props> = ({ className }: Props) => {
```
Nathan-Notifi, line 7: should we update this to take svg props too?

packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -115,6 +153,57 @@ input:-webkit-autofill {
...
+.NotifiAlertBox__content {
+  flex: 1;
+  display: flex;
+  margin-left: 8px;
+  margin-right: 8px;
+  color: var(--notifi-secondary-font-color);
+  font-weight: 400;
+  font-size: 15px;
+  line-height: 18px;
+  margin-bottom: 0 !important;
```
Nathan-Notifi, line 191: would it better if we remove the !last-child : mb-10? haha

packages/notifi-react-example/src/App.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 import './App.css';
 import { NotifiCard } from './NotifiCard/NotifiCard';
+import './Potem.css';
```
Nathan-Notifi, line 5: does this need to be in here?

packages/notifi-core/lib/operations/ConnectWallet.ts
juni-notifi: account ID is only required for certain chains!

packages/notifi-react-card/lib/assets/PenIcon.tsx
juni-notifi: 👍

packages/notifi-react-card/lib/components/defaults.css
juni-notifi: actually probably. I'm going to take a pass at it. I think instead of every decendant selector it should be immediate child selector

packages/notifi-react-example/src/App.tsx
juni-notifi: ah no, but it's a cool demo of what's possible with theming.

PR #183, opened 2023-02-20
updates alert labels from showing up in the preview view

Nathan-Notifi:
image

Labels would show up in the preview panel because they're technically an 'event type' from the config tool.


nimesh-notifi:
Is deposit confirmations on there twice expected?

hellonathanchung:
yeah, that's just from the config tool

juni-notifi (approved):

PR #186, opened 2023-02-21
Add support for balance alerts

juni-notifi:
- Consolidate logic in createAlert vs instantSubscribe
- Add support for walletBalance eventType
- Add support for AlertConfiguration with multiple sources
- Use connectedWallets to create sources

tested with multiwallettest dc08714b8b014ee79aec3ea650e34091

Screen Shot 2023-02-21 at 1 10 02 PM

Screen Shot 2023-02-21 at 1 10 13 PM

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
+++ b/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
@@ -2,7 +2,7 @@ import type {
   Alert,
   ClientData,
   ConnectWalletParams,
-  Filter,
+  CreateSourceInput,
   Source,
 } from '@notifi-network/notifi-core';
 import {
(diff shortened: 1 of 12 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -246,6 +246,155 @@ export const useNotifiSubscribe: ({
...
+          filterOptions,
+          filterType,
+          sources: sourcesInput,
+        } = alertConfiguration;
+        const sources = await Promise.all(sourcesInput.map(ensureSource));
+        const filter = sources
+          .flatMap((s) => s.applicableFilters)
+          .find((f) => f.filterType === filterType);
+        if (filter === undefined || filter.id === null) {
+          await deleteThisAlert();
```
Nathan-Notifi, line 304: is there a situation when we would need to delete the alert with a filter/source being undefined?

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -246,6 +246,155 @@ export const useNotifiSubscribe: ({
...
+          // Call serially because of limitations
+          await deleteThisAlert();
+          const alert = await client.createAlert({
+            emailAddress: finalEmail,
+            filterId: filter.id,
+            filterOptions: filterOptions ?? undefined,
+            groupName: 'managed',
+            name: alertName,
+            phoneNumber: finalPhoneNumber,
+            sourceId: '',
```
Nathan-Notifi, line 323: ah, so we're including `sourceId` and `sourceIds`?

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
juni-notifi: ah I'm just keeping the old behavior for now I think.

Nathan-Notifi (approved):

PR #189, opened 2023-02-25
updates health factor to handle prevent firing 

Nathan-Notifi:
limited the healfth factor subscriptions from being fired off if there is still a request in flight.

also updated the custom threshold configuration dividing things by 100 -- remnant from previous configurations

https://user-images.githubusercontent.com/105258726/221396371-67116e2a-117b-4b88-a43f-137b919ba5a5.mov


packages/notifi-react-card/lib/components/subscription/EventTypeCustomHealthCheckRow.tsx
```diff
@@ -299,6 +329,7 @@ export const EventTypeCustomHealthCheckRow: React.FC<
                 setErrorMessage('');
                 setSelectedIndex(null);
               }}
+              type="number"
               onBlur={handleCustomRatioButtonNewSubscription}
```
juni-notifi, line 333: should we prevent this handler from firing if `isNotificationLoading` is true as well?

juni-notifi: in fact, i would just `disable` this entire input if `isNotificationLoading` is true

Nathan-Notifi: good point

packages/notifi-react-card/lib/components/subscription/EventTypeBroadcastRow.tsx
```diff
@@ -46,30 +54,71 @@ export const EventTypeBroadcastRow: React.FC<EventTypeBroadcastRowProps> = ({
...
   useEffect(() => {
-    if (loading) {
+    if (didFetch.current) {
       return;
     }
+
     const hasAlert = alerts[alertName] !== undefined;
     setEnabled(hasAlert);
+    didFetch.current = true;
   }, [alertName, alerts, loading]);
```
juni-notifi, line 67: we can remove `loading` from dependencies i think

(and do we need to add the ref? technically i don't think we do)

packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
```diff
@@ -61,7 +61,7 @@ export const NotifiCard: React.FC = () => {
       >
         <NotifiSubscriptionCard
           darkMode
-          inputs={{ userWallet: publicKey }}
+          inputs={{ userWallet: '0xbC04dF9bc041f0aefdBC7DccBf7dA86717Caa1Ae' }}
```
juni-notifi, line 64: oop let's revert sample for now?

juni-notifi (approved):
Let's ship it and run thru some testing

packages/notifi-react-card/lib/components/subscription/EventTypeBroadcastRow.tsx
Nathan-Notifi: i think i had to ref to prevent really quick flickering on the toggles. let me remove and check again

Nathan-Notifi: yah, w/o the ref, it flickers

PR #191, opened 2023-02-27
updates loading state card css

Nathan-Notifi:
we had removed previous implementations of css here.

https://user-images.githubusercontent.com/105258726/221660396-5c094665-aee4-46d1-b7cf-dfe2d80d3a63.mov

### DIFF
--- a/packages/notifi-react-card/lib/components/common/LoadingStateCard.tsx
+++ b/packages/notifi-react-card/lib/components/common/LoadingStateCard.tsx
@@ -2,23 +2,32 @@ import clsx from 'clsx';
 import React from 'react';
 
 import { LoadingState } from '../../hooks';
+import Spinner from './Spinner';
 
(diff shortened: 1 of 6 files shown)


packages/notifi-react-card/lib/components/common/LoadingStateCard.tsx
```diff
@@ -2,23 +2,32 @@ import clsx from 'clsx';
...
   card: LoadingState;
   classNames?: Readonly<{
-    glimmer?: string;
+    container?: string;
+    label?: string;
   }>;
 }>;
 
 export const LoadingStateCard: React.FC<LoadingStateCardProps> = ({
   classNames,
```
marukohao, line 16: should we pass in the size and borderTopColor for Spinner here for customizing?

marukohao (approved):

PR #195, opened 2023-02-28
updates react card css to add progress spinner

Nathan-Notifi:
Previously we tried using css to update the Input but it's actually not visible,
instead we should add another class to the span when it's disabled that can be targeted.

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/NotifiToggle.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/NotifiToggle.tsx
@@ -36,11 +36,9 @@ export const NotifiToggle: React.FC<NotifiToggleProps> = ({
         }}
       />
       <span
-        className={clsx(
-          'NotifiToggle__slider',
-          classNames?.slider,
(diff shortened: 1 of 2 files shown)


juni-notifi (approved):

packages/notifi-react-card/lib/components/subscription/NotifiToggle.tsx
```diff
@@ -35,7 +35,13 @@ export const NotifiToggle: React.FC<NotifiToggleProps> = ({
           setChecked(e.target.checked);
         }}
       />
-      <span className={clsx('NotifiToggle__slider', classNames?.slider)}></span>
+      <span
+        className={clsx(
+          'NotifiToggle__slider',
+          classNames?.slider,
+          disabled ? 'NotifiToggleSlider--disabled' : undefined,
```
juni-notifi, line 42: alternative syntax:
```
{ 'NotifiToggleSlider--disabled': !disabled }
```
https://github.com/lukeed/clsx

juni-notifi (approved):

PR #199, opened 2023-03-03
Add fetchData function to frontendClient

juni-notifi:
As a proof-of-concept for future versions of the client, add a `fetchData` which fetches all of the data that are relevant to a user's state.


juni-notifi:
Closing without merging -- we may revisit this access pattern in the future.

PR #204, opened 2023-03-10
Make copy configurable from top level

juni-notifi:
We already had some fields which were configurable thru the frontend integration. Make some additional fields configurable (and fix some discovered bugs).

For now, we are just passing it through from the top level, but in the future we should consider pulling these values in from the fetched CardConfig.

### DIFF
--- a/packages/notifi-react-card/lib/components/common/ErrorStateCard.tsx
+++ b/packages/notifi-react-card/lib/components/common/ErrorStateCard.tsx
@@ -2,10 +2,14 @@ import clsx from 'clsx';
 import React from 'react';
 
 import { ErrorState } from '../../hooks';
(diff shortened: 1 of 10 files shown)


Nathan-Notifi:
Ooo nice!
I feel like we should remove the Get Notifications or hide it by default haha

image

Nathan-Notifi:
also i noticed the get notified doesnt looks slightly off

juni-notifi:
ah I think i fixed that by updating default style for AlertBox__button--spacer

Screen Shot 2023-03-09 at 5 31 53 PM


would you mind pulling again, if you're testing locally?

Nathan-Notifi (approved):

PR #206, opened 2023-03-10
Port PriceChange row

juni-notifi:
This row was implemented in the Vue integration of Pontem (and thus, in the notifi-frontend-client), but was never ported to the React SDK. Add support for this new row.


Nathan-Notifi (approved):
trueeeeee

PR #209, opened 2023-03-12
Refactor AlertDetails Markup

juni-notifi:
Before, we were rendering a full-sized card over the whole thing.

Instead, extract the state up to the SubscriptionCard and alter the rendering to be conditional on this state.

This drastically improves the flexibility, because it results in simpler, cleaner CSS and allows us to use the layout engine rather than having to pixel-push to match the container size.

Screen Shot 2023-03-12 at 1 44 53 PM

### DIFF
--- a/packages/notifi-react-card/lib/components/AlertHistory/AlertDetailsCard.tsx
+++ b/packages/notifi-react-card/lib/components/AlertHistory/AlertDetailsCard.tsx
@@ -2,15 +2,14 @@ import { NotificationHistoryEntry } from '@notifi-network/notifi-core';
 import clsx from 'clsx';
 import React, { useMemo } from 'react';
 
-import { BackArrowIcon } from '../../assets/BackArrowIcon';
 import {
   formatAlertDetailsTimestamp,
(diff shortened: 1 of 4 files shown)


Nathan-Notifi (approved):
WOW, huge! 👏

packages/notifi-react-card/lib/components/AlertHistory/AlertDetailsCard.tsx
```diff
@@ -2,23 +2,21 @@ import { NotificationHistoryEntry } from '@notifi-network/notifi-core';
 import clsx from 'clsx';
 import React, { useMemo } from 'react';
 
-import { BackArrowIcon } from '../../assets/BackArrowIcon';
 import {
   formatAlertDetailsTimestamp,
   formatAmount,
 } from '../../utils/AlertHistoryUtils';
+import NotifiAlertBox from '../NotifiAlertBox';
```
happyeric77, line 9: Seems the NotifiAlertBox is not used here

PR #211, opened 2023-03-14
Support Optimism on SDK

mayuranganesathas:
Not to be merged till Optimism mainnet approval

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiContext.tsx
@@ -54,9 +54,8 @@ export type AcalaParams = Readonly<{
 
 export type OptimismParams = Readonly<{
   walletBlockchain: 'OPTIMISM';
-  accountAddress: string;
   walletPublicKey: string;
-  signMessage: AcalaSignMessageFunction;
+  signMessage: Uint8SignMessageFunction;
 }>;
 
 
--- a/packages/notifi-react-card/lib/context/NotifiContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiContext.tsx
@@ -52,6 +52,14 @@ export type AcalaParams = Readonly<{
   signMessage: AcalaSignMessageFunction;
 }>;
 
+export type OptimismParams = Readonly<{
+  walletBlockchain: 'OPTIMISM';
+  accountAddress: string;
+  walletPublicKey: string;
+  signMessage: AcalaSignMessageFunction;
+}>;
+
+
 export type NearParams = Readonly<{
   walletBlockchain: 'NEAR';
   accountAddress: string;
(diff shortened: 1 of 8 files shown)


packages/notifi-react-card/lib/context/NotifiContext.tsx
```diff
@@ -52,6 +52,14 @@ export type AcalaParams = Readonly<{
   signMessage: AcalaSignMessageFunction;
 }>;
 
+export type OptimismParams = Readonly<{
+  walletBlockchain: 'OPTIMISM';
+  accountAddress: string;
```
juni-notifi, line 57: maybe i'm missing sometning, but other EVM params don't require accountAddress -- I think we should copy/paste an EVM param payload instead.

Better: can you take a stab and DRYing up the EVM params? it'll probably work if we do something like
```
export type EvmParams = Readonly<{
  walletBlockchain: 'ETHEREUM' | 'POLYGON' | ... | 'OPTIMISM';
  ...
}>
```

I don't know what ramifications that would have on the rest of the codebase though, so if it's too complicated just abort

packages/notifi-react-card/lib/context/NotifiContext.tsx
```diff
@@ -52,6 +52,14 @@ export type AcalaParams = Readonly<{
...
+  walletPublicKey: string;
+  signMessage: AcalaSignMessageFunction;
```
juni-notifi, line 59: why AcalaSignMessage instead of `Uint8SignMessageFunction`, which is used by other EVM stuff?

juni-notifi (changes requested):
I think we copy/pasta'd the wrong block in a few places.

This should be evident when you run `npx lerna run build` -- I'm like 72% sure that you'll see typescript compile errors.

Would you mind updating the commented areas and making sure that build succeeds?

packages/notifi-react-card/lib/context/NotifiContext.tsx
mayuranganesathas: sorry - fat finger. Resolving.

packages/notifi-react-card/lib/context/NotifiContext.tsx
mayuranganesathas: can give it a whirl - whats the appropriate approach to testing the sdk?

Nathan-Notifi: it's been fairly manual. 
i think if you can logIn with the wallet via the notifi-react-example that would be good

Nathan-Notifi (approved):

packages/notifi-react-card/lib/context/NotifiContext.tsx
mayuranganesathas: Creating backlog ticket -- will be an issue in the future. Have to clean up referencing for Wallet Params as Notifi-Core seems like the source of truth.

juni-notifi (approved):

Nathan-Notifi:
hey @mayuranganesathas!
is this something you'd want to merge?

PR #212, opened 2023-03-15
Move link from Logo to Footer

juni-notifi:
For some reason we had put the anchor tag inside the logo, which makes it so that only the filled parts of the SVG are clickable. Lift the link up to contain both the txt and the svg


Nathan-Notifi (approved):
🚀

PR #219, opened 2023-03-16
Add fallback renderer/ Filter out unsupported notification (MVP-2289)

eric-notifi:
# MVP-2289: Add fallback renderers for EventDetails

Add a new hook, useNotificationHistory, for the following 3 purposes:

1. Filter out the unsupported detailEntryType so that unsupported Entry will not be shown on historyCard. 
> *[Recommended approach](https://virtuoso.dev/troubleshooting/) from Virtuoso:*
> ```
> If you have zero-height items, you need to filter those out before passing them to the component. 
> The internal data structure needs a distinct position for each item. There's no way to fix this.
> ```

2. Add the fallback renderer (But this should never happen):
Screenshot 2023-03-16 at 13 23 49

3. Consolidate all the custom EventDetailRenderer components: we can get the associated props to pass into AlertNotificationRow by using this hook.

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
@@ -11,7 +11,11 @@ import { useNotifiClientContext } from '../../../context';
 import { DeepPartialReadonly } from '../../../utils';
 import { MESSAGES_PER_PAGE } from '../../../utils/constants';
 import { AccountBalanceChangedRenderer } from '../../AlertHistory/AccountBalanceChangedRenderer';
-import { AlertNotificationViewProps } from '../../AlertHistory/AlertNotificationRow';
+import { AlertIcon } from '../../AlertHistory/AlertIcon';
+import {
+  AlertNotificationRow,
+  AlertNotificationViewProps,
+} from '../../AlertHistory/AlertNotificationRow';
 import { BroadcastMessageChangedRenderer } from '../../AlertHistory/BroadcastMessageChangedRenderer';
 import { ChatMessageReceivedRenderer } from '../../AlertHistory/ChatMessageReceivedRenderer';
 import { GenericDetailRenderer } from '../../AlertHistory/GenericDetailRenderer';
@@ -115,8 +119,20 @@ export const AlertCard: React.FC<{
           walletAddress={notification.sourceAddress ?? ''}
         />
       );
+    default:
+      // It should never come here because exception should be filtered out before. https://virtuoso.dev/troubleshooting/
+      return (
+        <AlertNotificationRow
+          handleAlertEntrySelection={handleAlertEntrySelection}
+          notificationTitle={'Unsupported notification'}
+          notificationImage={<AlertIcon icon={'INFO'} />}
+          notificationSubject={'Alert not supported yet'}
+          notificationDate={''}
+          notificationMessage={''}
+          classNames={classNames}
+        />
+      );
   }
-  return null;
 };
 
 export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
@@ -176,6 +192,22 @@ export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
     }
   }, [client]);
 
+  const validateSupportedDetailType = useCallback(
+    (entry?: NotificationHistoryEntry): boolean => {
+      const supportedEntryDetailTypes = [
+        'BroadcastMessageEventDetails',
+        'HealthValueOverThresholdEventDetails',
+        'GenericEventDetails',
+        'ChatMessageReceivedEventDetails',
+        'AccountBalanceChangedEventDetails',
+      ];
+      if (supportedEntryDetailTypes.includes(entry?.detail?.__typename ?? ''))
+        return true;
+      return false;
+    },
+    [],
+  );
+
   return (
     <div
       className={clsx(
@@ -202,9 +234,7 @@ export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
               });
             }
           }}
-          data={allNodes.filter(
-            (notification) => notification.detail != undefined,
-          )}
+          data={allNodes.filter(validateSupportedDetailType)}
           itemContent={(_index, notification) => {
             return (
               <AlertCard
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -176,6 +192,22 @@ export const AlertHistoryView: React.FC<AlertHistoryViewProps> = ({
...
+        'HealthValueOverThresholdEventDetails',
+        'GenericEventDetails',
+        'ChatMessageReceivedEventDetails',
+        'AccountBalanceChangedEventDetails',
+      ];
+      if (supportedEntryDetailTypes.includes(entry?.detail?.__typename ?? ''))
+        return true;
+      return false;
+    },
+    [],
```
juni-notifi, line 208: a `useCallback` with no dependencies can simply be a module-level function -- let's move the function declaration outside the component (just moving it to line 137 should be sufficient)

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -115,8 +119,20 @@ export const AlertCard: React.FC<{
           walletAddress={notification.sourceAddress ?? ''}
         />
       );
+    default:
+      // It should never come here because exception should be filtered out before. https://virtuoso.dev/troubleshooting/
```
juni-notifi, line 123: 👍

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -115,8 +119,20 @@ export const AlertCard: React.FC<{
...
+      return (
+        <AlertNotificationRow
+          handleAlertEntrySelection={handleAlertEntrySelection}
+          notificationTitle={'Unsupported notification'}
+          notificationImage={<AlertIcon icon={'INFO'} />}
+          notificationSubject={'Alert not supported yet'}
+          notificationDate={''}
```
juni-notifi, line 130: not super important, but we can still use `notification.createdDate` probably

juni-notifi (approved):
LGTM

Only fragile piece is that we have to keep the array in-sync with the switch statement. Any good ideas for improvement?

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
eric-notifi: Refactored quit a bit, all related logics are migrated to useNotificationHistory hook.

packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
eric-notifi: Migrated to useNotificationHistory hook with some refactor.

packages/notifi-react-card/lib/hooks/useNotificationHistory.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+import { AnnouncementIcon } from '../assets/AnnouncementIcon';
+import { ChatAlertIcon } from '../assets/ChatAlertIcon';
+import { RatioCheckIcon } from '../assets/RatioCheckIcon';
+import { SwapIcon } from '../assets/SwapIcon';
+import { AlertIcon } from '../components/AlertHistory/AlertIcon';
+import { AlertNotificationViewProps } from '../components/AlertHistory/AlertNotificationRow';
+import { formatAmount } from '../utils/AlertHistoryUtils';
+
+export const useNotificationHistory = () => {
```
juni-notifi, line 19: since the hook doesn't actually consume any other hook dependencies, I don't think this should be a hook.

declaring another `.ts` module inside utils is probably cleaner

packages/notifi-react-card/lib/hooks/useNotificationHistory.tsx
```diff
@@ -0,0 +1,149 @@
...
+      generateProps: (
+        notification: NotificationHistoryEntry,
+      ) => AlertNotificationViewProps,
+    ) {
+      this.typename = typename;
+      this.generateProps = generateProps;
+    }
+  }
+
+  const supportedHistoryDetails: SupportedEntryDetail[] = [
```
juni-notifi, line 36: it would be better to see this as a Record<string, SupportedEntryDetail | undefined> so that `.find` is O[1]

packages/notifi-react-card/lib/hooks/useNotificationHistory.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { AnnouncementIcon } from '../assets/AnnouncementIcon';
+import { ChatAlertIcon } from '../assets/ChatAlertIcon';
+import { RatioCheckIcon } from '../assets/RatioCheckIcon';
+import { SwapIcon } from '../assets/SwapIcon';
+import { AlertIcon } from '../components/AlertHistory/AlertIcon';
+import { AlertNotificationViewProps } from '../components/AlertHistory/AlertNotificationRow';
+import { formatAmount } from '../utils/AlertHistoryUtils';
+
+export const useNotificationHistory = () => {
+  class SupportedEntryDetail {
```
juni-notifi, line 20: what's the benefit to this having a constructor? I think just an interface or a `type` declaration is sufficient

packages/notifi-react-card/lib/hooks/useNotificationHistory.tsx
```diff
@@ -0,0 +1,149 @@
...
+        };
+      },
+    ),
+  ];
+
+  const validateIsSupported = (entry?: NotificationHistoryEntry): boolean => {
+    if (
+      supportedHistoryDetails
+        .map((d) => d.typename)
+        .includes(entry?.detail?.__typename ?? '')
```
juni-notifi, line 121: the function `validateIsSupported` is run many times a frame -- let's consider improving its performance so we don't:
1. create a new list via .map
2. iterate over it (O[n]) via .includes

since the number of supportedHistoryDetails is fairly low, I think it's sufficient to construct a Set<string> beforehand (e.g. at the module level)

juni-notifi (changes requested):
I've left some minor feedback -- can you take another pass at it?

Actually I think we could merge the previous code first and then make this improvement as a separate PR, what do you think?

eric-notifi:
Branch has reset back the original commit. 
The syncing between switch statements and supportedEventDetails Will be done in a new extra PR .

juni-notifi (approved):

juni-notifi (approved):

PR #222, opened 2023-03-17
MVP-2281: Update styling of destination container

eric-notifi:
Add destination container.

Screenshot 2023-03-17 at 16 57 01

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -30,14 +30,18 @@
   --notifi-focus-color: #fff;
   --notifi-toggle-handle: #fff;
   --notifi-toggle-inactive: rgba(255, 255, 255, 0.1);
-  --notifi-toggle-active: rgba(255, 255, 255, 0.4);
-  --notifi-tooltip-background: rgba(0, 0, 0, 0.9);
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -30,14 +30,18 @@
   --notifi-focus-color: #fff;
   --notifi-toggle-handle: #fff;
   --notifi-toggle-inactive: rgba(255, 255, 255, 0.1);
-  --notifi-toggle-active: rgba(255, 255, 255, 0.4);
-  --notifi-tooltip-background: rgba(0, 0, 0, 0.9);
+  --notifi-toggle-active: linear-gradient(
+    207.53deg,
+    #fe7970 11.91%,
+    #feb776 88.99%
+  );
```
juni-notifi, line 37: fancy -- did we have a figma for this?

juni-notifi (commented):
The screenshot shows an underline on the Telegram "verify" link -- is that a regression?

Regardless, let's put a text-decoration: none

eric-notifi:
I believe we will not need the underline anywhere in the card. So I disabled it globally.

Nathan-Notifi (approved):

PR #223, opened 2023-03-18
Upgrade React in build

juni-notifi:
Use version 18 as peer dependency in our dev builds.

In the process, I found that the latest version of graphql-codegen was improperly using a transitive dependency. Resolved it by adding an explicit dependency on an older version of the graphql-request library


Nathan-Notifi (approved):

eric-notifi (approved):
Working as expected.

PR #225, opened 2023-03-20
Make Alert history title / description take up two rows (MVP-2286)

eric-notifi:
If title or description is long, let it take up to two rows. See the following example of Direct Push notification alert.
## Before:
Screenshot 2023-03-20 at 13 56 38

## After:
Screenshot 2023-03-20 at 13 48 51


Nathan-Notifi (approved):
LGTM! @eric-notifi

PR #228, opened 2023-03-24
MVP-2257: Support previewing view by inputing demoByJSONConfig arg 

eric-notifi:
Have NotifiSubscriptionCard support one more arg `demoByJSONConfig` to show the demo preview. See the the following.

https://user-images.githubusercontent.com/127958634/227394005-369d9696-c9dc-44aa-83df-7a3a12d9831c.mov

### DIFF
--- a/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
@@ -33,7 +33,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   intercomEmailInputContainerStyle,
   intercomView,
 }: NotifiEmailInputProps) => {
-  const { emailIdThatNeedsConfirmation, intercomCardView } =
+  const { emailIdThatNeedsConfirmation, intercomCardView, demoPreview } =
     useNotifiSubscriptionContext();
 
   const {
@@ -48,9 +48,11 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
 
   const { email: emailErrorMessage } = formErrorMessages;
 
-  const { resendEmailVerificationLink } = useNotifiSubscribe({
-    targetGroupName: 'Intercom',
-  });
+  const subscribe = demoPreview
+    ? null
+    : useNotifiSubscribe({
+        targetGroupName: 'Intercom',
+      });
 
   const validateEmail = () => {
     if (email === '') {
@@ -68,7 +70,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   };
 
   const handleClick = () => {
-    resendEmailVerificationLink();
+    subscribe?.resendEmailVerificationLink();
   };
 
   return (
--- a/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
+++ b/packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
@@ -33,7 +33,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   intercomEmailInputContainerStyle,
   intercomView,
 }: NotifiEmailInputProps) => {
-  const { emailIdThatNeedsConfirmation, intercomCardView, demoPreview } =
+  const { emailIdThatNeedsConfirmation, intercomCardView } =
     useNotifiSubscriptionContext();
 
   const {
@@ -48,11 +48,9 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
 
   const { email: emailErrorMessage } = formErrorMessages;
 
-  const subscribe = demoPreview
-    ? null
-    : useNotifiSubscribe({
-        targetGroupName: 'Intercom',
-      });
+  const { resendEmailVerificationLink } = useNotifiSubscribe({
+    targetGroupName: 'Intercom',
+  });
 
   const validateEmail = () => {
     if (email === '') {
@@ -70,7 +68,7 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
   };
 
   const handleClick = () => {
-    subscribe?.resendEmailVerificationLink();
+    resendEmailVerificationLink();
   };
 
   return (
--- a/packages/notifi-react-card/lib/components/WalletList/WalletList.tsx
+++ b/packages/notifi-react-card/lib/components/WalletList/WalletList.tsx
@@ -49,9 +49,11 @@ export const WalletListInternal: React.FC<WalletListInternalProps> = ({
   );
 };
 export const WalletList: React.FC = () => {
-  const { params } = useNotifiClientContext();
+  const {
+    params: { multiWallet },
+  } = useNotifiClientContext();
   const { connectedWallets, loading } = useNotifiSubscriptionContext();
-  const owned = params.multiWallet?.ownedWallets;
+  const owned = multiWallet?.ownedWallets;
 
   return (
     <WalletListInternal
--- a/packages/notifi-react-card/lib/components/WalletList/WalletList.tsx
+++ b/packages/notifi-react-card/lib/components/WalletList/WalletList.tsx
@@ -21,34 +21,42 @@ export const WalletListInternal: React.FC<WalletListInternalProps> = ({
   ownedWallets,
   disabled,
 }: WalletListInternalProps) => {
+  const { demoPreview } = useNotifiSubscriptionContext();
   return (
     <ul>
-      {ownedWallets.map((wallet) => {
-        return (
-          <ConnectWalletRow
-            key={wallet.walletPublicKey}
-            {...wallet}
-            connectedWallets={connectedWallets}
-            disabled={disabled}
-          />
-        );
-      })}
+      {!demoPreview ? (
+        ownedWallets.map((wallet) => {
+          return (
+            <ConnectWalletRow
+              key={wallet.walletPublicKey}
+              {...wallet}
+              connectedWallets={connectedWallets}
+              disabled={disabled}
+            />
+          );
+        })
+      ) : (
+        <ConnectWalletRow
+          key="Demo view"
+          walletBlockchain="SOLANA"
+          walletPublicKey="0x0notifi"
+          signMessage={async (msg) => msg}
+          connectedWallets={connectedWallets}
+          disabled={disabled}
+        />
+      )}
     </ul>
   );
 };
 export const WalletList: React.FC = () => {
-  const {
-    params: { multiWallet },
-  } = useNotifiClientContext();
+  const { demoPreview } = useNotifiSubscriptionContext();
+  const clientContext = demoPreview ? null : useNotifiClientContext();
   const { connectedWallets, loading } = useNotifiSubscriptionContext();
-  const owned = multiWallet?.ownedWallets;
-  if (owned === undefined || owned.length === 0) {
-    return null;
-  }
+  const owned = clientContext?.params.multiWallet?.ownedWallets;
 
   return (
     <WalletListInternal
-      ownedWallets={owned}
+      ownedWallets={owned ?? []}
       connectedWallets={connectedWallets}
       disabled={loading}
     />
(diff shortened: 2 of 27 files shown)


juni-notifi (commented):
Oh, instead of a separate parameter for `demoPreview`, is it possible to create a mock Subscription Context instead?

packages/notifi-react-card/lib/components/NotifiEmailInput.tsx
```diff
@@ -48,9 +48,11 @@ export const NotifiEmailInput: React.FC<NotifiEmailInputProps> = ({
...
   const { email: emailErrorMessage } = formErrorMessages;
 
-  const { resendEmailVerificationLink } = useNotifiSubscribe({
-    targetGroupName: 'Intercom',
-  });
+  const subscribe = demoPreview
+    ? null
+    : useNotifiSubscribe({
+        targetGroupName: 'Intercom',
+      });
```
juni-notifi, line 55: hooks cannot be rendered conditionally, otherwise React gets angry -- can we make `useNotifiSubscribe` depend on the Subscription Context and handle the network issues in there?

packages/notifi-react-card/lib/components/WalletList/WalletList.tsx
```diff
@@ -21,34 +21,42 @@ export const WalletListInternal: React.FC<WalletListInternalProps> = ({
...
+      )}
     </ul>
   );
 };
 export const WalletList: React.FC = () => {
-  const {
-    params: { multiWallet },
-  } = useNotifiClientContext();
+  const { demoPreview } = useNotifiSubscriptionContext();
+  const clientContext = demoPreview ? null : useNotifiClientContext();
```
juni-notifi, line 53: same comment re: conditional hook rendering

eric-notifi:
Issue addressed. Avoid conditional call of hook and move the preview check logic into hook.

packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCardContainer.tsx
```diff
@@ -23,15 +23,24 @@ export const NotifiSubscriptionCardContainer: React.FC<
...
   onClose,
+  demoByJSONConfig,
 }: React.PropsWithChildren<NotifiSubscriptionCardProps>) => {
   const { isInitialized } = useNotifiSubscribe({ targetGroupName: 'Default' });
-  const { loading } = useNotifiSubscriptionContext();
+  const { loading, setDemoPreview } = useNotifiSubscriptionContext();
+
+  useEffect(() => {
+    demoByJSONConfig ? setDemoPreview(demoByJSONConfig) : setDemoPreview(null);
+  }, [demoByJSONConfig]);
```
juni-notifi, line 33: curious -- any reason we have to set this via useEffect? why can't we just pass it as a prop to the NotifiSubscriptionContext?

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -45,6 +55,8 @@ export type NotifiSubscriptionData = Readonly<{
   setConversationId: (conversationId: string) => void;
   userId: string;
   setUserId: (userId: string) => void;
+  demoPreview: DemoPreview | null;
+  setDemoPreview: React.Dispatch<React.SetStateAction<DemoPreview | null>>;
```
juni-notifi, line 59: I think setting this via prop solely via prop feels cleanest

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -78,6 +90,10 @@ export const NotifiSubscriptionContextProvider: React.FC<
   const [email, setEmail] = useState<string>('');
   const [phoneNumber, setPhoneNumber] = useState<string>('');
   const [telegramId, setTelegramId] = useState<string>('');
+  const [demoPreview, setDemoPreview] = useState<DemoPreview | null>({
+    view: 'expired',
+    data: defaultDemoConfigV1,
+  });
```
juni-notifi, line 96: i'd feel better if the default value is null

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -204,11 +206,19 @@ export const useNotifiSubscribe: ({
         .catch((_e) => {
           /* Intentionally empty */
         });
+    } else if (demoPreview) {
+      // Mockup info for demo preview card
+      setEmail(defaultDemoConfigV1.name);
+      setPhoneNumber('+101234567890');
+      setTelegramId(defaultDemoConfigV1.id!);
```
juni-notifi, line 213: let's just have an early return if `demoPreview` is not null, and rely on the initial values passed to the demoPreview

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -204,11 +206,19 @@ export const useNotifiSubscribe: ({
...
     }
   }, [client.isAuthenticated]);
 
   const logInViaHardwareWallet =
     useCallback(async (): Promise<SubscriptionData> => {
+      if (demoPreview) {
+        return dummySubscribeData;
```
juni-notifi, line 220: feel free to throw an error instead

edit: same with all the other methods

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -624,13 +640,23 @@ export const useNotifiSubscribe: ({
   return {
     resendEmailVerificationLink,
     instantSubscribe,
-    isAuthenticated: client.isAuthenticated,
-    isInitialized: client.isInitialized,
-    isTokenExpired: client.isTokenExpired,
+    isAuthenticated: client?.isAuthenticated ?? false,
+    isInitialized: client?.isInitialized ?? false,
+    isTokenExpired: client?.isTokenExpired ?? false,
```
juni-notifi, line 645: we no longer need the null check on client, right?

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -624,13 +640,23 @@ export const useNotifiSubscribe: ({
...
+
+const dummySubscribeData: SubscriptionData = {
+  alerts: { dummy: {} as unknown as Alert },
+  email: null,
+  phoneNumber: null,
+  telegramId: null,
+  telegramConfirmationUrl: null,
+  isPhoneNumberConfirmed: null,
+  emailIdThatNeedsConfirmation: '',
+};
```
juni-notifi, line 662: in general, having "dummy" data spread across many files is an architectural smell, ideally we can have it in one place (like a DummySubscriptionContext or something) and have others cleanly consuming

maybe it's time to re-think the dependencies between useNotifiSubscribe, useNotifiClient, etc. etc....

packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCardContainer.tsx
```diff
@@ -23,15 +23,24 @@ export const NotifiSubscriptionCardContainer: React.FC<
...
-  const card = useSubscriptionCard({
-    id: cardId,
-    type: 'SUBSCRIPTION_CARD',
-  });
+  const card = useSubscriptionCard(
+    {
+      id: cardId,
+      type: 'SUBSCRIPTION_CARD',
+    },
+    demoByJSONConfig,
```
juni-notifi, line 42: instead of parameterizing useSubscriptionCard this way, I think we can mount a different component (maybe don't mount a NotifiSubscriptionCardContainer?) if we are using demo view.

packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
```diff
@@ -90,13 +95,165 @@ export const NotifiCard: React.FC = () => {
...
+          cardId=""
+          demoByJSONConfig={{ view: 'expired', data: JSON.parse(data) }}
+        />
+        Dummy Demo NotifiSubscriptionCard: Verify page
+        <NotifiSubscriptionCard
+          darkMode
+          cardId=""
+          demoByJSONConfig={{ view: 'verify', data: JSON.parse(data) }}
+        />
+      </NotifiContext>
```
juni-notifi, line 139: ohhh, I now see why we wanted to structure things this way. This does feel pretty "clean" because at first sight the "demo config" modifies the subscription card.

However, I think lifting the demoByJSONConfig up to `NotifiContext` might be cleaner (since we don't want it to have any functionality when we're in demo mode)

juni-notifi (commented):
Would you mind taking another pass, and seeing if you can limit the number of inner components that need to change?

Ideally we show almost no lines of diff in most components

Let me know if you want to find time to pair on this, that might be the most productive (to reduce iteration time)

packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCardContainer.tsx
eric-notifi: Migrated demoPreview to newly created 'NotifiDemoPreviewContext' that handles demoPreview related context

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
eric-notifi: Already Migrated demoPreview to newly created 'NotifiDemoPreviewContext' that handles demoPreview related context.So this section no longer exists.

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
eric-notifi: Already Migrated demoPreview to newly created 'NotifiDemoPreviewContext' that handles demoPreview related context which contain new default value as null.

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
eric-notifi: You are right. addressed.

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
eric-notifi: We no longer need the `dummySubscribeData` because we throw error instead of returning dummy value. 
And Already Migrated demoPreview to newly created 'NotifiDemoPreviewContext' that includes related demoPreview consts and types.

packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
eric-notifi: Very appreciate this hint, it is now moved up to `NotifiContext`

eric-notifi:
# Refactor note:
1. Instead of passing demoPreview data from `NotifiSubscriptionCard`, now using `NotifiContext` to do so.
2. Instead of handling demoPreview related logic everywhere, centralising these into new context `NotifiDemoPreviewContext`.
3. Make sure all the card components not changing.

> ## Overview :
> Screenshot 2023-03-27 at 13 52 06

packages/notifi-react-example/src/NotifiCard/NotifiCard.tsx
```diff
@@ -60,6 +60,10 @@ export const NotifiCard: React.FC = () => {
   };
 
   return (
+    /**
+     * Test dappAdresses/ subscription cardId / Intercom cardId:
+     * junitest.xyz / b07f0267df1343db9d14757c653ec092 / 1045f61752b148eabab0403c08cd60b2
```
Nathan-Notifi, line 65: do we need these comments here?

there is a chance we delete these ids in the future

eric-notifi: Thanks for the head-up. comments removed.

juni-notifi (approved):

Nathan-Notifi:
@eric-notifi hey eric did you mean to merge and close?

PR #229, opened 2023-03-24
Error message state for destinations.

Nathan-Notifi:
With tooltip content and error message.

in the future, can update this more as the information from the backend becomes more solidified
image

on hover:
image

without tooltip
image

for testing:
- manually update the subscription context original state


packages/notifi-react-card/lib/context/constants.tsx
```diff
@@ -0,0 +1,12 @@
+export type DestinationInputs = {
+  email: string;
+  phoneNumber: string;
+  telegram: string;
+};
```
juni-notifi, line 5: Record<FormField, string>

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -138,6 +140,10 @@ export const useNotifiSubscribe: ({
 
       setIsSmsConfirmed(isPhoneNumberConfirmed);
 
+      if (!isPhoneNumberConfirmed) {
+        setPhoneNumberErrorMessage('Invalid number');
+      }
```
juni-notifi, line 145: seeing this makes me feel like there's a lot of undefined state that's giong to be tough to maintain. How hard would it be to re-design the `confirmed` state to use the new "destination error" state?

spitballing off top of my head, but maybe something like
```
type RecoverableError = Readonly<{
  type: 'recoverableError';
  label: string;
  onClick: () => void;
}>;

type UnrecoverableError = Readonly<{
  type: 'unrecoverableError';
  label: string;
  tooltip?: string;
}>;

type DestinationError = UnrecoverableError | RecoverableError;
type DestinationErrors = Record<FormField, DestinationError>;
```

juni-notifi (commented):
Would you mind seeing if we can consolidate the "unverified" state into the same system?

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
Nathan-Notifi: will take a look!

packages/notifi-react-card/lib/context/NotifiFormContext.tsx
```diff
@@ -24,18 +25,13 @@ export type NotifiFormData = Readonly<{
   setTelegramErrorMessage: (value: string) => void;
```
marukohao, line 25: Should the type of the value be DestinationError here?

Nathan-Notifi: for now, no. this is in the formContext so it's handling it a little more simply compared to the subscription context

PR #230, opened 2023-03-29
MVP-2257 [Rebased Version]: Support previewing view by inputing demoByJSONConfig arg

eric-notifi:
# Change note
1. MVP-2257 with rebased changes
2. Add NotifiDemoPreivewProvider on top lever that allow users to use demo preview card
3. Change structure:
Screenshot 2023-03-29 at 13 06 17

4. Demo example

https://user-images.githubusercontent.com/127958634/228425679-15858696-aaaa-46e1-b29a-18e07551cb57.mov

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useSubscriptionCard.ts
+++ b/packages/notifi-react-card/lib/hooks/useSubscriptionCard.ts
@@ -2,9 +2,8 @@ import { ClientFetchSubscriptionCardInput } from '@notifi-network/notifi-core';
 import { useEffect, useState } from 'react';
 
 import {
-  DemoPreview,
-  defaultDemoConfigV1,
   useNotifiClientContext,
+  useNotifiDemoPreviewContext,
 } from '../context';
 import { CardConfigItemV1 } from './SubscriptionCardConfig';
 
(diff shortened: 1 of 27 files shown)


packages/notifi-react-card/lib/hooks/useSubscriptionCard.ts
```diff
@@ -61,5 +64,14 @@ export const useSubscriptionCard = (
       });
   }, [input.id, input.type]);
 
+  useEffect(() => {
+    if (demoPreview && state.state === 'error') {
+      setState(() => ({
+        state: 'fetched',
+        data: demoPreview.data,
+      }));
+    }
```
juni-notifi, line 73: let's move this into the above useEffect, before doing anything else

juni-notifi (approved):

packages/notifi-react-card/lib/hooks/useSubscriptionCard.ts
eric-notifi: IMO, It might not be possible to move it to the above useEffect since the 'client.fetchSubscriptionCard' is an async func that makes the useEffect to loop infinitely if we add the  `state` as dependency (state.state change between loading and fetched) .

eric-notifi: Combined to only one useEffect.

PR #232, opened 2023-03-30
Fix linting issues caused by lack of PropsWithChildren

eric-notifi:
# Problem
Linting error when building
```
lerna ERR! npm run build stderr:
lib/components/intercom/NotifiIntercomCard.tsx(46,8): error TS2559: Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'.
```
# Fix
## Add PropsWithChildren to related components.


Nathan-Notifi (approved):
🚀

PR #233, opened 2023-04-01
Discord updates

Nathan-Notifi:
This is a doozy.

**So far,** this should allow users to toggle on for Discord and then generate a confirmation url that opens in a new window.

When toggled off, it should hide the discord field in the user info panel.
If a user has no discord targets, a random one should be created.

If active, the verify link should be be hidden.


packages/notifi-core/lib/operations/CreateDiscordTarget.ts
```diff
@@ -0,0 +1,28 @@
+import { DiscordTarget, Operation } from '../models';
+
+/**
+ * Input param for creating an DiscordTarget
+ *
+ * @remarks
+ * This describes the values to use for email target
```
eric-notifi, line 7: Might want to change it to `Discord` target

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -237,6 +290,7 @@ export const useNotifiSubscribe: ({
       // Mockup info for demo preview card
       setEmail(defaultDemoConfigV1.name);
       setPhoneNumber('+101234567890');
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       setTelegramId(defaultDemoConfigV1.id!);
```
eric-notifi, line 294: Adding the dummy discord for preview card might be a good idea: `setDiscrodId(defaultDemoConfigV1.id!)`

eric-notifi (commented):
LGTM, Just left few small points.

packages/notifi-core/lib/operations/CreateDiscordTarget.ts
Nathan-Notifi: great catch

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
Nathan-Notifi: yeah, i added a dummy code for the telegram id so avoid the nonnull assertion as well

eric-notifi (approved):
It looks good to me.

packages/notifi-axios-adapter/lib/queries/getDiscordTargetsImpl.ts
```diff
@@ -0,0 +1,30 @@
...
+  discordTargetFragment,
+  discordTargetFragmentDependencies,
+} from '../fragments';
+
+const DEPENDENCIES = [
+  ...discordTargetFragmentDependencies,
+  discordTargetFragment,
+];
+
+const MUTATION = `
```
eric-notifi, line 17: Not sure if it should be called `QUERY` ?

Nathan-Notifi: you're right. that is confusing. let me do a sweep and check

PR #241, opened 2023-04-07
Discord fixes

Nathan-Notifi:
Updated infinite render loop issue caused by listening to missing targets data and setDiscordTarget triggering each other.

also added tooltip to update text for sms if they've typed 'stop'

### DIFF
--- a/packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
+++ b/packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
@@ -225,7 +225,6 @@ const useNotifiClient = (
     isInitialized: boolean;
     isTokenExpired: boolean;
     expiry: string | null;
-    reenableSmsPhoneNumber: string;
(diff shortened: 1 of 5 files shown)


packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
```diff
@@ -247,6 +248,10 @@ const useNotifiClient = (
   const [isAuthenticated, setIsAuthenticated] = useState<boolean>(false);
   const [expiry, setExpiry] = useState<string | null>(null);
   const [isTokenExpired, setisTokenExpired] = useState<boolean>(false);
+
+  const reenableSmsPhoneNumber =
+    env === 'Production' ? '+1 206 222 3465' : '+1 253 880 1477 ';
```
juni-notifi, line 253: do we have to expose this from the client? why not have these two values in the same file as the error?

juni-notifi (approved):

PR #250, opened 2023-04-13
remove uuid

Nathan-Notifi:
uuid doesnt work with vite. fortunately, we shouldnt need to pass in a randomly generated id. in the subscribe method, we create one with the name default anyways.

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
+++ b/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
@@ -9,7 +9,6 @@ import {
 } from '@notifi-network/notifi-core';
 import { isValidPhoneNumber } from 'libphonenumber-js';
 import { useCallback, useEffect, useRef, useState } from 'react';
-import { v4 as uuid } from 'uuid';
 
 import {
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -134,7 +133,7 @@ export const useNotifiSubscribe: ({
     // If there is a target, set the discord target data to it, otherwise create a new target.
     setDiscordTargetData(
```
juni-notifi, line 134: I think i mentioned this previously, but I think it's probably better if we make this a `useState<DiscordTarget | undefined>` and use `undefined` value to have the semantic meaning of "no target available" rather than having a placeholder with fake data.

juni-notifi (approved):

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
Nathan-Notifi: ooo i have to address that change. let me update the logic

PR #253, opened 2023-04-14
Error state

Nathan-Notifi:
image


packages/notifi-react-card/lib/hooks/useFetchedCardState.ts
```diff
@@ -22,14 +22,20 @@ export type VerifyOnboardingViewState = Readonly<{
   state: 'verifyonboarding';
 }>;
 
+export type ErrorViewState = Readonly<{
+  state: 'error';
+  reason: unknown;
```
juni-notifi, line 27: Ah, I guess this is reusing the same pattern as the "fetch" state error card

I think the error card just renders the `reason` as a string -- it would be nice if we can update the logic in ErrorCard a little bit so we can format the errors a little bit more nicely.

juni-notifi (approved):

packages/notifi-react-card/lib/hooks/useFetchedCardState.ts
Nathan-Notifi: ill do it now

PR #254, opened 2023-04-17
adds loading state card for edit card

Nathan-Notifi:
There was no loading indicator to the user when submitting the card.

also, because we had `logIn` setting loading to false, it'd prematurely flicker back to the card.
https://user-images.githubusercontent.com/105258726/232606571-edf84b91-2041-4f7e-a680-974efd4c39ec.mov

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiDemoPreviewContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiDemoPreviewContext.tsx
@@ -19,7 +19,7 @@ export type DemoPreview = {
 export const defaultDemoConfigV1: CardConfigItemV1 = {
   version: 'v1',
   id: '@notifi.network', // Shown as dummy telegram id
-  name: 'notofi@notifi.network', // Shown as dummy email field
+  name: 'notifi@notifi.network', // Shown as dummy email field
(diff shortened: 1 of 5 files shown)


packages/notifi-react-card/lib/context/NotifiDemoPreviewContext.tsx
```diff
@@ -19,7 +19,7 @@ export type DemoPreview = {
 export const defaultDemoConfigV1: CardConfigItemV1 = {
   version: 'v1',
   id: '@notifi.network', // Shown as dummy telegram id
-  name: 'notofi@notifi.network', // Shown as dummy email field
+  name: 'notifi@notifi.network', // Shown as dummy email field
```
juni-notifi, line 22: oh, it'd be nice if we can move this into notifi-react-example somehow, instead of in notifi-react-card

juni-notifi (approved):

packages/notifi-react-card/lib/context/NotifiDemoPreviewContext.tsx
Nathan-Notifi: oh absolutely

PR #264, opened 2023-04-24
MVP-2512: Implement userState getter and subscribeWallet method in NotifiFrontendClient

eric-notifi:
1. Instead of copy-pasting type `SignMessageParams`, import it from `@notifi-network/notifi-core`

2. Implement a getter to get the userState. It is the replacement of the following 4 states in useNotifiClient (Check the detail in [SPIKE DOC](https://docs.colorfullife.ml/notifi/sdk-consolidate#step1-implement-getter-to-get-the-userstate)):

- isAuthenticated
- isInitialized
- expiry
- isTokenExpired

3. Implement `subscribeWallet` method in `NotifiFrontendClient` class. It is the replacement of original `subscribeWallet` method in `useNotifiSubscribe`

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -1,3 +1,7 @@
+import {
+  ConnectWalletParams,
+  SignMessageParams,
+} from '@notifi-network/notifi-core';
 import { Types } from '@notifi-network/notifi-graphql';
 import { NotifiService } from '@notifi-network/notifi-graphql';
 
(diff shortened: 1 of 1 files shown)


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -1,3 +1,7 @@
+import {
+  ConnectWalletParams,
+  SignMessageParams,
+} from '@notifi-network/notifi-core';
```
juni-notifi, line 4: ah actually I wanted to deprecate `notifi-core` since it's a little uncomfortable to use.

The ideal developer experience should be that I can import all the types I need from the package that I depend on (i.e. if I'm only using react-card, I should be able to import everything from it)

For now, I think it's okay because it helps us to DRY things

juni-notifi (approved):

PR #266, opened 2023-04-25
MVP-2513: Implement fetchData method in NotifiFrontendClient

eric-notifi:
1. Implement fetchData to NotifiFrontendClient (This will be replacing the fetching data method in `useNotifiClient` in notifi-react-hooks)

2. Showcase `fetchData` method and `useState` getter in react-example. See the demo video below:
  

https://user-images.githubusercontent.com/127958634/234154114-63797871-69fc-401c-a88c-a53ffcb0a15c.mov

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -22,76 +22,6 @@ import {
   ensureWebhook,
 } from './ensureTarget';
 
-export type Alert = NonNullable<
-  NonNullable<Types.GetAlertsQuery['alert']> extends Array<infer T> ? T : never
->;
-
-export type ConnectedWallet = NonNullable<
-  NonNullable<Types.GetConnectedWalletsQuery['connectedWallet']> extends Array<
-    infer T
-  >
-    ? T
-    : never
->;
-
-export type EmailTarget = NonNullable<
-  NonNullable<Types.GetEmailTargetsQuery['emailTarget']> extends Array<infer T>
-    ? T
-    : never
->;
-
-export type Filter = NonNullable<
-  NonNullable<Types.GetFiltersQuery['filter']> extends Array<infer T>
-    ? T
-    : never
->;
-
-export type SmsTarget = NonNullable<
-  NonNullable<Types.GetSmsTargetsQuery['smsTarget']> extends Array<infer T>
-    ? T
-    : never
->;
-
-export type Source = NonNullable<
-  NonNullable<Types.GetSourcesQuery['source']> extends Array<infer T>
-    ? T
-    : never
->;
-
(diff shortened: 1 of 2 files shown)


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -280,6 +350,65 @@ export class NotifiFrontendClient {
...
+    ] = await Promise.all([
+      this._service.getAlerts({}),
+      this._service.getConnectedWallets({}),
+      this._service.getSources({}),
+      this._service.getTargetGroups({}),
+      this._service.getEmailTargets({}),
+      this._service.getSmsTargets({}),
+      this._service.getTelegramTargets({}),
+      this._service.getDiscordTargets({}),
+      this._service.getFilters({}),
```
juni-notifi, line 373: ah, this pattern was due to an old server-side bug which prevented us from querying multiple fields at once.

Insetad, let's define a `service.fetchData` which makes a composite query

i.e. something like:
```
query FetchData {
  alerts { ...AlertFragment }
  connectedWallets { ... }
  sources { .... }
  ...
}

File: packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts [hunk 2]
```diff
@@ -22,6 +22,76 @@ import {
...
+  alerts: ReadonlyArray<Alert>;
+  connectedWallets: ReadonlyArray<ConnectedWallet>;
+  emailTargets: ReadonlyArray<EmailTarget>;
+  filters: ReadonlyArray<Filter>;
+  smsTargets: ReadonlyArray<SmsTarget>;
+  sources: ReadonlyArray<Source>;
+  targetGroups: ReadonlyArray<TargetGroup>;
+  telegramTargets: ReadonlyArray<TelegramTarget>;
+  discordTargets: ReadonlyArray<DiscordTarget>;
+};
```
juni-notifi, line 93: if we add a separate `fetchData` query, then we can simply expose the generated GraphQL type directly rather than having to re-declare the types here.

juni-notifi (changes requested):
Can you look into using a composite query instead of composing queries and re-declaring the types?

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: Addressed! Export the `FetchDataQuery` as `FrontendClientData`

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: Addressed!

juni-notifi (approved):

PR #272, opened 2023-04-27
MVP-2533: Add NotifiAcalaConfiguration and related methods in notifi-frontend-client

eric-notifi:
1. newAcalaClient
2. condition check to add prefix in storage driver (memory & indexDB)
3. ACALA _signMessage/login method in NotifiFrontendClient
4. newAcalaConfig

### DIFF
--- a/packages/notifi-frontend-client/lib/storage/InMemoryStorageDriver.ts
+++ b/packages/notifi-frontend-client/lib/storage/InMemoryStorageDriver.ts
@@ -33,6 +33,7 @@ export const createInMemoryStorageDriver = (
       break;
     }
     case 'SUI':
+    case 'ACALA':
     case 'APTOS': {
       keyPrefix += `:${config.accountAddress}:${config.authenticationKey}`;
       break;
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -34,6 +34,11 @@ export type AptosSignMessageFunction = (
   message: string,
   nonce: number,
 ) => Promise<string>;
+type hexString = `0x${string}`;
+export type AcalaSignMessageFunction = (
+  acalaAddress: string,
+  message: string,
+) => Promise<hexString>;
 
 type BeginLoginProps = Omit<Types.BeginLogInByTransactionInput, 'dappAddress'>;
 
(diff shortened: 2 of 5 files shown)


packages/notifi-frontend-client/lib/storage/InMemoryStorageDriver.ts
```diff
@@ -32,6 +32,7 @@ export const createInMemoryStorageDriver = (
       keyPrefix += `:${config.walletPublicKey}`;
       break;
     }
+    case 'ACALA':
```
eric-notifi, line 35: In react-hooks, the implementation does not have chain specific storage prefix. All is set as `${jwtPrefix}:${dappAddress}:${walletPublicKey}`. But in some chains, the publickey is not constant. 
I am wondering if it will be good idea to use `${jwtPrefix}:${dappAddress}:${accountAddress}` for those chains that have accountAddress.

```suggestion
    case 'ACALA':
    case 'APTOS': {
      keyPrefix += `:${config.accountAddress}`;
      break;
    }
```

packages/notifi-frontend-client/lib/storage/InMemoryStorageDriver.ts
```diff
@@ -32,6 +32,7 @@ export const createInMemoryStorageDriver = (
...
     case 'APTOS': {
       keyPrefix += `:${config.accountAddress}:${config.authenticationKey}`;
       break;
```
eric-notifi, line 38: In react-hooks, the implementation [does not have chain specific storage prefix](https://github.com/notifi-network/notifi-sdk-ts/blob/main/packages/notifi-react-hooks/lib/utils/storage.ts#L32). All is set as `${jwtPrefix}:${dappAddress}:${walletPublicKey}`. But in some chains, the publickey is not constant. 
I am wondering if it will be good idea to use `${jwtPrefix}:${dappAddress}:${accountAddress}` for those chains that have accountAddress.

```suggestion
    case 'ACALA':
    case 'APTOS': {
      keyPrefix += `:${config.accountAddress}`;
      break;
    }
```

Nathan-Notifi:
would something like this be possible too?

```
    if (
      signMessageParams.walletBlockchain !==
      this._configuration.walletBlockchain
    ) {
      throw new Error(
        'Sign message params and configuration must have the same blockchain',
      );
    }
```

so before we enter the switch statements, we can throw the error

Nathan-Notifi (approved):

eric-notifi:
Thanks! already make a note in MVP-2537

PR #273, opened 2023-04-27
MVP-2534: Add NotifiSuiConfiguration and related methods in notifi-frontend-client

eric-notifi:
1. newSuiClient
2. condition check to add prefix in storage driver (memory & indexDB) 
3. SUI _signMessage/login method in NotifiFrontendClient
4. newSuiConfig

### DIFF
--- a/packages/notifi-frontend-client/lib/client/clientFactory.ts
+++ b/packages/notifi-frontend-client/lib/client/clientFactory.ts
@@ -6,6 +6,7 @@ import {
   NotifiEvmConfiguration,
   NotifiFrontendConfiguration,
   NotifiSolanaConfiguration,
+  NotifiSuiConfiguration,
   envUrl,
 } from '../configuration';
 import {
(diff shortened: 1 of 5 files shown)


packages/notifi-frontend-client/lib/storage/InMemoryStorageDriver.ts
```diff
@@ -32,6 +32,7 @@ export const createInMemoryStorageDriver = (
       keyPrefix += `:${config.walletPublicKey}`;
       break;
     }
+    case 'SUI':
```
eric-notifi, line 35: ***CONCERN***
In this case, the accountAddress is the same as publicKey(authenticationKey). Not sure if it is correct to put in this group? or it would belong to the following group:
```
// ...
case 'ETHEREUM':
    case 'POLYGON':
    case 'ARBITRUM':
    case 'AVALANCHE':
    case 'BINANCE':
    case 'OPTIMISM':
    case 'SOLANA': {
      keyPrefix += `:${config.walletPublicKey}`;
      break;
    }
  // ...
```

packages/notifi-frontend-client/lib/client/clientFactory.ts
```diff
@@ -42,3 +43,9 @@ export const newEvmClient = (config: NotifiEvmConfiguration) => {
   const storage = newNotifiStorage(config);
   return new NotifiFrontendClient(config, service, storage);
 };
```
Nathan-Notifi, line 45: for these client factories,
is possible to have a type that inclues something like this?

type NotifiConfiguration =
  | NotifiAptosConfiguration
  | NotifiInjectiveConfiguration
  | NotifiSolanaConfiguration
  | NotifiEvmConfiguration;


and then try to generate a new frontendClient?

Nathan-Notifi (approved):

packages/notifi-frontend-client/lib/client/clientFactory.ts
eric-notifi: Nice point, will implement it in MVP-2537

PR #280, opened 2023-05-01
MVP-2555: Optimize frontendClient storage

eric-notifi:
1. DRY storage code

2. Add configure newNotifiStorage -- default to localForage, but have an option for in-memory.

### DIFF
--- a/packages/notifi-frontend-client/lib/client/clientFactory.ts
+++ b/packages/notifi-frontend-client/lib/client/clientFactory.ts
@@ -11,9 +11,9 @@ import { NotifiFrontendClient } from './NotifiFrontendClient';
 
 export const newNotifiStorage = (config: NotifiFrontendConfiguration) => {
   const driver =
-    config.storageOption === 'LocalForage'
-      ? createLocalForageStorageDriver(config)
-      : createInMemoryStorageDriver(config);
+    config.storageOption?.driverType === 'InMemory'
+      ? createInMemoryStorageDriver(config)
+      : createLocalForageStorageDriver(config);
   return new NotifiFrontendStorage(driver);
 };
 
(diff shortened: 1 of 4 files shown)


packages/notifi-frontend-client/lib/client/clientFactory.ts
```diff
@@ -4,12 +4,16 @@ import { GraphQLClient } from 'graphql-request';
...
   NotifiFrontendStorage,
+  createInMemoryStorageDriver,
   createLocalForageStorageDriver,
 } from '../storage';
 import { NotifiFrontendClient } from './NotifiFrontendClient';
 
 export const newNotifiStorage = (config: NotifiFrontendConfiguration) => {
-  const driver = createLocalForageStorageDriver(config);
+  const driver =
+    config.storageOption === 'LocalForage'
```
juni-notifi, line 14: would you mind inverting this?

i.e. `(storageOption === 'InMemory') createInMemoryStorage : createLocalForageStorage`

that way it undefined values will default to the previous default

packages/notifi-frontend-client/lib/configuration/NotifiFrontendConfiguration.ts
```diff
@@ -7,6 +7,7 @@ export type NotifiEnvironment =
 export type NotifiEnvironmentConfiguration = Readonly<{
   env: NotifiEnvironment;
   tenantId: string;
+  storageOption: 'LocalForage' | 'InMemory';
```
juni-notifi, line 10: please make this an object, in case we want to add further configuration for storage. Also, let's make it optional

```
storageOption?: Readonly<{
  driverType?: 'LocalForage' | 'InMemory'
}>
```

packages/notifi-frontend-client/lib/client/clientFactory.ts
eric-notifi: You are right. It should be like that, thanks!

packages/notifi-frontend-client/lib/configuration/NotifiFrontendConfiguration.ts
eric-notifi: Nice suggestion. Thanks!

Nathan-Notifi (approved):

PR #281, opened 2023-05-01
MVP-2556: frontendClient object accessible in notifi-react-card

eric-notifi:
1. Make fronttendClient object accessible in NotifiClientContext: for now both client object will be available until frontendClient can completely replace hooks' client.
2. Replace old `isInitalized` method in `NotifiSubscriptionCardContainer` using frontendClient

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiClientContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiClientContext.tsx
@@ -2,7 +2,6 @@ import {
   ConfigFactoryInput,
   NotifiFrontendClient,
   newFrontendClient,
-  newFrontendConfig,
 } from '@notifi-network/notifi-frontend-client';
 import { useNotifiClient } from '@notifi-network/notifi-react-hooks';
 import React, { createContext, useContext, useEffect, useMemo } from 'react';
@@ -53,8 +52,7 @@ export const NotifiClientContextProvider: React.FC<NotifiParams> = ({
 
   const frontendClient = useMemo(() => {
     const configInput = getFrontendConfigInput(params);
-    const config = newFrontendConfig(configInput);
-    return newFrontendClient(config);
+    return newFrontendClient(configInput);
   }, [
     params.dappAddress,
     params.env,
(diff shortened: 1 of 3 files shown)


packages/notifi-react-card/lib/context/NotifiClientContext.tsx
```diff
@@ -1,25 +1,73 @@
...
+  useEffect(() => {
+    // Init frontend client
+    if (!frontendClient.userState) {
+      frontendClient.initialize();
+    }
+  }, [frontendClient.userState?.status]);
+
   return (
-    <NotifiClientContext.Provider value={{ client, params }}>
+    <NotifiClientContext.Provider value={{ client, params, frontendClient }}>
```
eric-notifi, line 70: The `client` from `hooks` will be here until it is able to be completely replaced with `frontendClient` .

packages/notifi-react-card/lib/context/NotifiClientContext.tsx
```diff
@@ -1,25 +1,73 @@
...
+    params.walletBlockchain,
+    params.walletPublicKey,
+  ]);
+
+  useEffect(() => {
+    // Init frontend client
+    if (!frontendClient.userState) {
+      frontendClient.initialize();
+    }
+  }, [frontendClient.userState?.status]);
```
juni-notifi, line 67: I think it would be fine to re-render whenever `frontendClient` changes

```
useEffect(() => {
  frontendClient.initialize();
}, [frontendClient]);
```

Am I missing something?

packages/notifi-react-card/lib/context/NotifiClientContext.tsx
juni-notifi: If possible, it would be good to be able to use either hooks or frontendclient, so it will be easy to test the two flows with one card config

juni-notifi (approved):

packages/notifi-react-card/lib/context/NotifiClientContext.tsx
eric-notifi: Addressed!

eric-notifi:
Wrap the `frontendClient with something like canary object:
```
canary: {
    isActive: boolean;
    frontendClient: NotifiFrontendClient;
  };
  ```
It allows turning on and off frontendClient

PR #282, opened 2023-05-01
MVP-2558: HealthCheckEventTypeItem event type and its related methods in FrontendClient

eric-notifi:
Implement missing HealthCheckEventTypeItem type and its related helper functions that will be consumed by notifi-react-card


Nathan-Notifi (approved):
Looks good. @juni-notifi and i previously spoke about trying to move everything to using the customEventType later to reduce code.

maybe in a future PR we can try to do that!

PR #288, opened 2023-05-04
MVP-2577: 'frontendClient' object available in SubscriptionCardV1 component (notifi-react-card)

eric-notifi:
Subscription adopts frontendClient's logic when enableCanary is true. Otherwise, keeps original flow

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
@@ -122,9 +122,9 @@ export const SubscriptionCardV1: React.FC<SubscriptionCardV1Props> = ({
   const { isClientInitialized, isClientTokenExpired } = useMemo(() => {
     let isClientInitialized = false;
     let isClientTokenExpired = false;
-    if (canaryIsActive && frontendClient.userState) {
-      isClientInitialized = true;
-      isClientTokenExpired = frontendClient.userState.status === 'expired';
+    if (canaryIsActive) {
+      isClientInitialized = !!frontendClient.userState;
+      isClientTokenExpired = frontendClient.userState?.status === 'expired';
     } else {
       isClientInitialized = isInitialized;
       isClientTokenExpired = isTokenExpired;
--- a/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
@@ -10,6 +10,7 @@ import React, {
 } from 'react';
 
 import {
+  useNotifiClientContext,
   useNotifiDemoPreviewContext,
   useNotifiForm,
   useNotifiSubscriptionContext,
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
```diff
@@ -114,14 +115,34 @@ export const SubscriptionCardV1: React.FC<SubscriptionCardV1Props> = ({
...
+    let isClientTokenExpired = false;
+    if (canaryIsActive && frontendClient.userState) {
+      isClientInitialized = true;
+      isClientTokenExpired = frontendClient.userState.status === 'expired';
+    } else {
+      isClientInitialized = isInitialized;
+      isClientTokenExpired = isTokenExpired;
+    }
+    return { isClientInitialized, isClientTokenExpired };
+  }, [
```
Nathan-Notifi, line 133:
```suggestion
  const { isClientInitialized, isClientTokenExpired } = useMemo(() => {
    let isClientInitialized = canaryIsActive && frontendClient.userState) ?? false;
    let isClientTokenExpired = false;
    if (isClientInitialized) {
      isClientTokenExpired = frontendClient.userState.status === 'expired';
    } else {
      isClientInitialized = isInitialized;
      isClientTokenExpired = isTokenExpired;
    }
    
    wdyt?
    return { isClientInitialized, isClientTokenExpired };
  }, [
```

Nathan-Notifi (approved):

packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
eric-notifi: ooo, This is a perfect suggestion. One small think I realized when I tried it in action. The typescript has some compile error because that `isClientInitialized` is assigned as `null | boolean` to which the `isInitialized` is not able to be assigned. So I might want to change it to
```suggestion
  const { isClientInitialized, isClientTokenExpired } = useMemo(() => {
    let isClientInitialized =
      !!(canaryIsActive && frontendClient.userState) ?? false;
    let isClientTokenExpired = false;
    if (isClientInitialized) {
      isClientTokenExpired = frontendClient.userState?.status === 'expired';
    } else {
      isClientInitialized = isInitialized;
      isClientTokenExpired = isTokenExpired;
    }

    return { isClientInitialized, isClientTokenExpired };
  }
```

eric-notifi: Ah ... I think it might not be a good idea. 
By the original approaches, the `isClientInitialized` and `isClientTokenExpired` will be anyway fallback to adopt `useNotifiSubscribe` state even though `isCanary` is set as `true`. I believe we might go for 
```suggestion
 const { isClientInitialized, isClientTokenExpired } = useMemo(() => {
    let isClientInitialized = false;
    let isClientTokenExpired = false;
    if (canaryIsActive) {
      isClientInitialized = !!frontendClient.userState;
      isClientTokenExpired = frontendClient.userState?.status === 'expired';
    } else {
      isClientInitialized = isInitialized;
      isClientTokenExpired = isTokenExpired;
    }
    return { isClientInitialized, isClientTokenExpired };
  }, [
```

PR #292, opened 2023-05-05
MVP-2585-patch: re-render data after subscribe/ unsubscribe alter

eric-notifi:
1. Expose render method from `NotifiSubscriptionContext`
2. render after alter is subscribed / unsubscribed

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
@@ -81,6 +81,7 @@ export type NotifiSubscriptionData = Readonly<{
   setDiscordTargetData: React.Dispatch<
     React.SetStateAction<DiscordTarget | undefined>
   >;
+  render: (newData: Types.FetchDataQuery) => SubscriptionData;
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -199,6 +201,17 @@ export const NotifiSubscriptionContextProvider: React.FC<
...
+    frontendClient
+      .fetchData()
+      .then((data) => {
+        render(data);
+      })
+      .catch((_e) => {
+        throw new Error('error re-rendering data');
+      });
+  }, [reRenderTriggered]);
+
```
juni-notifi, line 214: hm, this feels like a code smell for a missing `render` somewhere -- I think instead of exposing an additional param, it would be better to plug the gap where we have the issue.

eric-notifi: Addressed. Exposing useCallback instead of useState + useEffect

juni-notifi (approved):

PR #300, opened 2023-05-09
MVP-2603: 'subscribeAlert'  in 'EventTypePriceChangeRow' 

eric-notifi:
Implement 'subscribeAlert/ unsubscrigbeAlert' in EventTypePriceChangeRow that allowed client to subscribe alert using either hooks or frontendClient

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/EventTypePriceChangeRow.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/EventTypePriceChangeRow.tsx
@@ -44,7 +44,12 @@ export type EventTypePriceChangeRowProps = Readonly<{
 
 export const EventTypePriceChangeRow: React.FC<
   EventTypePriceChangeRowProps
-> = ({ classNames, disabled, config }: EventTypePriceChangeRowProps) => {
+> = ({
+  classNames,
+  disabled,
+  config,
+  inputs,
+}: EventTypePriceChangeRowProps) => {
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/subscription/EventTypePriceChangeRow.tsx
```diff
@@ -69,16 +123,17 @@ export const EventTypePriceChangeRow: React.FC<
 
     if (!enabled) {
       setEnabled(true);
-      instantSubscribe({
-        alertConfiguration: alertConfiguration,
-        alertName: alertName,
+      subscribeAlert({
+        eventType: config,
+        inputs: {},
```
juni-notifi, line 128: is it okay that `inputs` is always an empty object?

juni-notifi (approved):

packages/notifi-react-card/lib/components/subscription/EventTypePriceChangeRow.tsx
eric-notifi: in [`ensurePriceChangeSource`](https://github.com/notifi-network/notifi-sdk-ts/blob/b6881792f87ffecb8713ce83aa4a23eaecb2853f/packages/notifi-frontend-client/lib/client/ensureSource.ts#L113), it does not make use use of `inputs` to `PriceChangeEventTypeItem`. So the inputs arg is always `{}`.

PR #301, opened 2023-05-09
MVP-2609-patch: 'getCustomFilterOptions' fallback value (frontend-client)

eric-notifi:
- Add Fallback val in 'getCustomFilterOptions' allows customFilterOption able to fallback to checkRatios[0] if client does not pass in either `threshold` or `value`

- This also helps when dapp wants to ensure multiple alerts at the same time (alternative method of `subscribe` in `useNotifiSubscribe`

**NOTE**
Blocker of MVP-2610

### DIFF
--- a/packages/notifi-frontend-client/lib/client/ensureSource.ts
+++ b/packages/notifi-frontend-client/lib/client/ensureSource.ts
@@ -615,13 +615,18 @@ const getCustomFilterOptions = (
           { type: 'ref', ref: thresholdDirectionKey },
           inputs,
         ) ?? eventType.checkRatios[0].ratio; // Fallback to 1st checkRatios
+
+      if (!healthRatio || !thresholdDirection) {
+        throw new Error('Failed to retrieve health ratio or direction');
+      }
+
       return {
         alertFrequency: eventType.alertFrequency,
         threshold:
           eventType.numberType === 'percentage'
             ? healthRatio / 100
             : healthRatio,
-        thresholdDirection: thresholdDirection === 'below' ? 'below' : 'above',
+        thresholdDirection: thresholdDirection === 'above' ? 'above' : 'below',
       };
     }
   }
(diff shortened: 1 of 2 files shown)


packages/notifi-frontend-client/lib/client/ensureSource.ts
```diff
@@ -599,21 +599,29 @@ const getCustomFilterOptions = (
...
-        { type: 'ref', ref: healthRatioKey },
-        inputs,
-      );
-
+      const healthRatio =
+        resolveNumberRef(
+          healthRatioKey,
+          { type: 'ref', ref: healthRatioKey },
+          inputs,
+        ) ?? eventType.checkRatios[0].ratio; // Fallback to 1st checkRatios
```
juni-notifi, line 609: what if `checkRatios` is an empty array?

packages/notifi-frontend-client/lib/client/ensureSource.ts
```diff
@@ -599,21 +599,29 @@ const getCustomFilterOptions = (
...
       return {
         alertFrequency: eventType.alertFrequency,
         threshold:
           eventType.numberType === 'percentage'
             ? healthRatio / 100
             : healthRatio,
-        thresholdDirection: eventType.checkRatios[0]?.type ?? 'below',
+        thresholdDirection: thresholdDirection === 'below' ? 'below' : 'above',
```
juni-notifi, line 624: this changes the default value I think

```suggestion
        thresholdDirection: thresholdDirection === 'above' ? 'above' : 'below',
```

juni-notifi (approved):

packages/notifi-frontend-client/lib/client/ensureSource.ts
eric-notifi: Addressed! Add checkRatios length 0 case
```
if (!healthRatio || !thresholdDirection) {
    throw new Error('Failed to retrieve health ratio or direction');
}
```

packages/notifi-frontend-client/lib/client/ensureSource.ts
eric-notifi: Addressed!! Thanks!

PR #307, opened 2023-05-17
we want to have conditional contexts.

Nathan-Notifi:
this is primarily to allow users higher level contexts that the subscription card consumes. without it, we might have weird interactions with multiple instances of contexts existing.


juni-notifi (approved):

packages/notifi-react-card/lib/context/NotifiSubscriptionContext.tsx
```diff
@@ -94,13 +96,15 @@ const hasKey = <K extends string>(
 ): obj is object & { [k in K]: unknown } => {
   return typeof obj === 'object' && obj !== null && key in obj;
 };
+const contextId = uuid();
```
juni-notifi, line 99: probably more useful if we do this inside the Provider render function as a `useMemo`

juni-notifi (approved):

PR #310, opened 2023-05-18
MVP-2659: Remove notifi-core dependency from notifi-frontendClient

eric-notifi:
Remove notifi-core imports from notifi-frontendClient. Migrate the following types to `notifi-frontend-client` so that it no longer depends on notifi-core (will be deprecated later):
  - SignMessageParams
  - WalletWithSignParams
  - WalletWithSignMessage
  - ConnectWalletParams


Nathan-Notifi (approved):
noice!

PR #311, opened 2023-05-18
MVP-2660: Consolidate targetGroup types in notifi-core and notifi-graphql

eric-notifi:
In order to being able to completely deprecate notifi-core in notifi-react-card, firstly we need to consolidate types of notifi-core and notifi-graphql (targets).
Accordingly, also adjust corresponding parts

### DIFF
--- a/packages/notifi-core/lib/models/ConversationMessages.ts
+++ b/packages/notifi-core/lib/models/ConversationMessages.ts
@@ -1,3 +1,5 @@
+import { Types } from '@notifi-network/notifi-graphql';
+
 export type ConversationMessages = Readonly<{
   nodes?: Array<ConversationMessagesEntry> | undefined;
   pageInfo: {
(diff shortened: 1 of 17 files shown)


packages/notifi-core/lib/models/ConversationMessages.ts
```diff
@@ -6,16 +8,9 @@ export type ConversationMessages = Readonly<{
...
-  __typename?: 'ConversationMessage';
-  conversationParticipant: Participant;
-  id: string;
-  message: string;
-  userId: string;
-  createdDate: string;
-  updatedDate: string;
-  conversationId: string;
-}>;
+export type ConversationMessagesEntry = NonNullable<
```
Nathan-Notifi, line 11: question! do we need the nonNullable around the nonNullable?

Nathan-Notifi (approved):

packages/notifi-core/lib/models/ConversationMessages.ts
eric-notifi: Nice catch!! it is the same. Thanks for pointing out. resolved.

PR #312, opened 2023-05-19
MVP-2660-patch: follow graphql-codegen convention using Fragment type

eric-notifi:
Optimization: Use the type with `Fragment` suffix instead of extracting the same types from `FetchQuery`


Nathan-Notifi (approved):
i like it

PR #313, opened 2023-05-19
MVP-2661: Remove notifi-core dependency from notifi-react-card

eric-notifi:
Remove the Type dependencies to `notifi-core` from `notifi-react-card`.


juni-notifi (commented):
Oh hmmmm some of these might not actually be compatible... can we wait on this until we can remove react-hooks entirely?

eric-notifi:
Np, let's deprecate these after removing hook.

PR #318, opened 2023-05-24
Fix Fusion Source Creation

juni-notifi:
The fusionEventTypeId needs to be fetched and compared.

We also need to resolve all the right config values in the frontend client


Nathan-Notifi (approved):

eric-notifi:
Cool, meaning from now all the sources will  necessarily have `fusionEvnetTypeId`. Shall we also need to add the filter to other ensureSourceType in forntendClient

PR #324, opened 2023-06-01
MVP-2744: Add fusionHealthCheck eventType

eric-notifi:
Add fusionHealthCheck event type

**NOTE**
Merge it after MVP-2745 since it changes the way how SDK resolve the config of fusion event type. The config generate by current AP is not able to work with this new approach.

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/86456317-99e7-42b0-89e6-e8fd27f162d4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
-**/node_modules
+**/node_modules*
+
 **/dist
 **/coverage
 **/.DS_Store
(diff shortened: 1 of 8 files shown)


Nathan-Notifi:
can we have it so a minimum of two health check ratios are displayed?

Nathan-Notifi (approved):

PR #328, opened 2023-06-07
MVP-2772: Consolidate NotifiSubscriptionContextProvider and NotifiFormPrvider

eric-notifi:
In some cases, we want to grant certain dapp page the access to consume the following contexts:

1. `NotifiSubscriptionContext` --> `useNotifiSubscriptionContext()`
2. `NotifiClientContext` --> `useNotifiClientContext()`

However, these two contexts are only available inside the `NotifiSubscriptionCard` component

This PR is to solve the problem. The scope is below:

- Move the `NotifiSubscriptionContextProvider` and `NotifiFormProvider` into `NotifiContext`

- Consolidate `NotifiSubscriptionCard` and `NotifiSubscriptionCardContainer`

- Implement `SolanaCard` example case

**DEMO VIDEO**

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/c47c0106-d62a-484d-8401-8f6f848377f0


Nathan-Notifi:
Hey Eric, we actually have it so that a context isnt created if one can be detected.

```tsx
export const NotifiSubscriptionCard: React.FC<
  React.PropsWithChildren<NotifiSubscriptionCardProps>
> = (props: React.PropsWithChildren<NotifiSubscriptionCardProps>) => {
  const { params } = useNotifiClientContext();
  const { contextId } = useNotifiSubscriptionContext();
  // NOTE: This allows the card to be used with contexts outside of the card.
  // The context already exists at this point
  if (contextId) {
    return <NotifiSubscriptionCardContainer {...props} />;
  }

  return (
    <NotifiFormProvider>
      <NotifiSubscriptionContextProvider {...params}>
        <NotifiSubscriptionCardContainer {...props} />
      </NotifiSubscriptionContextProvider>
    </NotifiFormProvider>
  );
};
```

i do like the idea where we pass in children and contents though. can we separate them into two separate PR's? I'm still iffy on consolidating them but if you feel strongly, it's okay with me

wdyt @juni-notifi ?

eric-notifi:
@Nathan-Notifi Let me align a bit, what you mean might be: 
Only move `NotifiSubscriptionContext` and `NotifiFromContext` to `NotifiContext` w/o consolidating `NotifiSubscriptionCard` and `NotifiSubscriptionCardContener`. 

IMO, by this way, The `NotifiSubscriptionCard` will only as wrapper of `NotifiSubscriptionContext` as below

```

export const NotifiSubscriptionCard: React.FC<
  React.PropsWithChildren<NotifiSubscriptionCardProps>
> = (props: React.PropsWithChildren<NotifiSubscriptionCardProps>) => {
  const { contextId } = useNotifiSubscriptionContext();
  // NOTE: This allows the card to be used with contexts outside of the card.
  // The context already exists at this point
  if (contextId) {
    return <NotifiSubscriptionCardContainer {...props} />;
  }

  return (
    <>
      <NotifiSubscriptionCardContainer {...props} />
    </>
  );
};
```
And in this case, the contextId will not make any effect because the `NotifiSubscriptionContext` will have been moved upper layer than `NotifiSubscriptionCard`. It might look a bit wried for me.

Nathan-Notifi (approved):

PR #333, opened 2023-06-12
MVP-2804: Support EVMOS in SDK

eric-notifi:
Add EVMOS chain support in SDK.

**NOTE**
Only can be merged after BE supports EVMOS chain (sourceType: EVMOS_WALLET)

### DIFF
--- a/packages/notifi-core/lib/NotifiClient.ts
+++ b/packages/notifi-core/lib/NotifiClient.ts
@@ -275,7 +275,8 @@ export type SignMessageParams =
         | 'AVALANCHE'
         | 'BINANCE'
         | 'INJECTIVE'
-        | 'OPTIMISM';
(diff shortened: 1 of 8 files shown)


juni-notifi:
EVMOS is a little bit special in that consumers might be using a Cosmos wallet or an EVM wallet -- how should we differentiate those?

eric-notifi:
Ah ... I did not know EVMOS also allows users to connect with cosmos wallet. 
IMO, only supporting one address type (EVM addr) can reduce the complexity. In this case, we might need to make a constraint dapp to input evm user address. 
dapp might need to use the [@evmos/address-converter](https://www.npmjs.com/package/@evmos/address-converter) to convert the user addr among each addr types.
Would it be ideal? @juni-notifi

PR #337, opened 2023-06-16
Add blockchains

kai-notifi:
Add missing blockchains to sdk

### DIFF
--- a/packages/notifi-core/lib/models/WalletBlockchain.ts
+++ b/packages/notifi-core/lib/models/WalletBlockchain.ts
@@ -1,7 +1,6 @@
 export type WalletBlockchain =
   | 'SOLANA'
   | 'ETHEREUM'
-  | 'OFF_CHAIN'
   | 'AVALANCHE'
   | 'APTOS'
   | 'ACALA'
@@ -11,6 +10,4 @@ export type WalletBlockchain =
   | 'NEAR'
   | 'OPTIMISM'
   | 'INJECTIVE'
-  | 'SUI'
-  | 'OSMOSIS'
-  | 'EVMOS';
+  | 'SUI';
--- a/packages/notifi-core/lib/models/WalletBlockchain.ts
+++ b/packages/notifi-core/lib/models/WalletBlockchain.ts
@@ -1,6 +1,7 @@
 export type WalletBlockchain =
   | 'SOLANA'
   | 'ETHEREUM'
+  | 'OFF_CHAIN'
   | 'AVALANCHE'
   | 'APTOS'
   | 'ACALA'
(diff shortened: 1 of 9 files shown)


eric-notifi:
Hey @kai-notifi Kai, Just checked the PR. OSMOSIS, EVMOS and OFF_CHAIN are added

If it is tend to add chains support, I guess there are some more places to add. I also have one [PR for supporting EVMOS](https://github.com/notifi-network/notifi-sdk-ts/pull/333/files) FYI. But at that time (two days ago), BE still not yet supports the EVMOS wallet source type, also not certain about the signMessage

eric-notifi:
Ah.... and I think we do not need to increase the version manually. Lerna will handle the version when we publish it.

packages/notifi-core/lib/models/WalletBlockchain.ts
```diff
@@ -10,4 +11,6 @@ export type WalletBlockchain =
   | 'NEAR'
   | 'OPTIMISM'
   | 'INJECTIVE'
-  | 'SUI';
+  | 'SUI'
+  | 'OSMOSIS'
+  | 'EVMOS';
```
eric-notifi, line 16: It might be better not to include the unsupported chains:
- OFF_CHAIN
- EVMOS
- OSMOSIS

packages/notifi-node/lib/types/WalletBlockchain.ts
```diff
@@ -1 +1,16 @@
...
+  | 'ACALA'
+  | 'POLYGON'
+  | 'ARBITRUM'
+  | 'BINANCE'
+  | 'NEAR'
+  | 'OPTIMISM'
+  | 'INJECTIVE'
+  | 'SUI'
+  | 'OSMOSIS'
+  | 'EVMOS';
```
eric-notifi, line 16: It might be better not to include the unsupported chains:
- OFF_CHAIN
- EVMOS
- OSMOSIS

packages/notifi-node-sample/package.json
```diff
@@ -14,7 +14,7 @@
     "dev": "nodemon lib/index.ts"
   },
   "dependencies": {
-    "@notifi-network/notifi-node": "^0.73.2",
+    "@notifi-network/notifi-node": "^0.73.3",
```
eric-notifi, line 17: Ah ... we might not need to manually modify the version. Lerna will handle version control when we publish it.

eric-notifi: It adopts to all version related section below

eric-notifi (changes requested):

Haoqian-Zheng:
lgtm!

eric-notifi (approved):
LGTM, Thanks!

PR #338, opened 2023-06-19
MVP-2839: Fix confirm link (discord/ tg) not update after enabling bot

eric-notifi:
Now the discord confirmation link still remain the same even the users have already accepted the invitation. Like below.

Screenshot 2023-06-16 at 23 17 59

This PR is to fix this issue.

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCard.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCard.tsx
@@ -90,7 +90,9 @@ export const NotifiSubscriptionCard: React.FC<
   loadingSpinnerSize,
   onClose,
 }) => {
-  const { isInitialized } = useNotifiSubscribe({ targetGroupName: 'Default' });
+  const { isInitialized, reload, isAuthenticated } = useNotifiSubscribe({
+    targetGroupName: 'Default',
+  });
   const {
     canary: { isActive: canaryIsActive, frontendClient },
   } = useNotifiClientContext();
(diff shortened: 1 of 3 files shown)


packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCard.tsx
```diff
@@ -110,6 +112,16 @@ export const NotifiSubscriptionCard: React.FC<
...
+  window.onfocus = () => {
+    // Ensure target is up-to-date after user returns to tab from 3rd party verification site
+    if (!isClientInitialized || !isAuthenticated) return;
+
+    if (canaryIsActive) {
+      return frontendClient.fetchData().then(render);
+    }
+    reload();
+  };
+
```
juni-notifi, line 124: we should install side-effects only inside `useEffect` hooks

```
useEffect(() => {
  const handler = () => {
    if (!isClientInitialized || !isAuthenticated) {
      return;
    }
    if (canaryIsActive) {
      return frontendClient.fetchData().then(render);
    }
    reload();
  };
  
  window.addEventListener('focus', handler);
  return () => {
    window.removeEventListener('focus', handler);
  };
}, [isClientInitialized, isAuthenticated, canaryIsActive, frontendClient]);

juni-notifi (changes requested):
In the past, we had explicitly agreed that this was something we wouldn't support -- did that change recently?

packages/notifi-react-card/lib/components/subscription/NotifiSubscriptionCard.tsx
eric-notifi: Thanks! It is much better solution!!

juni-notifi (approved):

PR #339, opened 2023-06-21
MVP-2860: re-signup fails when discord target enabled

eric-notifi:
## Replicate method

1. Signup a new wallet

2. Clean the browser cache

3. get in card and signup view will prompt up (expected because browser cache is cleaned)

4. toggle on the discord and signup.

5. Gql error 505See the error video demo attached herewith

See error video demo attched

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/e56f788a-5f12-49f1-80f0-7341af1450ad

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
+++ b/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
@@ -608,7 +608,12 @@ export const useNotifiSubscribe: ({
       let finalDiscordId = undefined;
 
       if (useDiscord === true) {
-        finalDiscordId = await client.createDiscordTarget('Default');
+        try {
+          finalDiscordId = await client.createDiscordTarget('Default');
+        } catch (e) {
+          // catch the case that the 'Default' discord already created
+          finalDiscordId = data.discordTargets[0].id;
+        }
       }
 
       const newResults: Record<string, Alert> = {};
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -608,7 +608,12 @@ export const useNotifiSubscribe: ({
       let finalDiscordId = undefined;
 
       if (useDiscord === true) {
-        finalDiscordId = await client.createDiscordTarget('Default');
+        try {
+          finalDiscordId = await client.createDiscordTarget('Default');
+        } catch (e) {
+          // catch the case that the 'Default' discord already created
+          finalDiscordId = data.discordTargets[0].id;
+        }
```
juni-notifi, line 616: instead of a try/catch pattern, we should do a "search -> if not found then create" pattern

```
const existingDiscordTarget = data.discordTargets.find(target => target.name === 'Default');
let finalDiscordId: string | undefined = undefined;
if (existingDiscordTarget !== undefined) {
  finalDiscordId = existingDiscordTarget.id;
} else {
  const createdDiscordTarget = await client.createDiscordTarget('Default');
  finalDiscordId = createdDiscordTarget.id;
}
```

juni-notifi (commented):
I think in general we should avoid broad try/catch statements, since they can mask network errors incorrectly

packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
eric-notifi: Right!! it is much better solution. Thanks!

juni-notifi (approved):

PR #340, opened 2023-06-21
MVP-2842: fix FrontendClient Discord verification fails

eric-notifi:
## ISSUE
When using frontendClient,
1. it will create the different discordId name (`default`) than the one created by `react-hooks` (`Default`). We need to make them consistent.
2. same situation as #339 while `enableCanary={true}`.

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/NotifiSubscribeButton.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/NotifiSubscribeButton.tsx
@@ -5,6 +5,7 @@ import {
 } from '@notifi-network/notifi-frontend-client';
 import clsx from 'clsx';
 import { isValidPhoneNumber } from 'libphonenumber-js';
+import { TargetGroupFragmentFragment } from 'notifi-graphql/lib/gql/generated';
 import { formatTelegramForSubscription } from 'notifi-react-card/lib/utils/stringUtils';
 import React, { useCallback, useMemo } from 'react';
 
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/components/subscription/NotifiSubscribeButton.tsx
```diff
@@ -106,7 +107,16 @@ export const NotifiSubscribeButton: React.FC<NotifiSubscribeButtonProps> = ({
 
   const renewTargetGroups = useCallback(async () => {
     if (isCanaryActive) {
-      return frontendClient.ensureTargetGroup(targetGroup);
+      let result = {} as TargetGroupFragmentFragment;
+      try {
+        result = await frontendClient.ensureTargetGroup(targetGroup);
+      } catch (e) {
+        // catch the case that the 'Default' discord already created
```
juni-notifi, line 114: Can you clarify this comment?

Do you mean the TargetGroup? rather than 'discord'

packages/notifi-react-card/lib/components/subscription/NotifiSubscribeButton.tsx
```diff
@@ -106,7 +107,16 @@ export const NotifiSubscribeButton: React.FC<NotifiSubscribeButtonProps> = ({
...
+        const data = await frontendClient.fetchData();
+        if (!data.targetGroup?.[0]) throw e;
```
juni-notifi, line 116: instead of getting the first targetGroup, we should do a `find`

```
const defaultTargetGroup = data.targetGroup?.find(it => it.name === 'Default');
if (defaultTargetGroup === undefined) {
  throw e;
}

result = defaultTargetGroup;
```

PR #342, opened 2023-06-22
Add support for 'Osmosis' value

juni-notifi:
Osmosis is another Cosmos chain, similar to Injective


eric-notifi (approved):
LGTM!
Wondering if we also want to add this into `packages/notifi-node/lib/types/WalletBlockchain.ts`

PR #346, opened 2023-06-29
Support custom targetGroupName in ensureAlert

juni-notifi:
Allow passing in a custom TargetGroup name durint ensureAlert in NotifiFrontendClient codepath.


eric-notifi (approved):
🚀

PR #348, opened 2023-07-05
Add support for AlertFrequency in Fusion alerts

juni-notifi:
Not sure why we don't support this -- simple piping of things through


eric-notifi (approved):
Ah .... Right, it is missing. Thanks!

PR #349, opened 2023-07-06
MVP-2762: Regular users being not able to re-direct to historyView 

eric-notifi:
# 🛠️ Regular users being not able to re-direct to historyView

The regular users (Not FTU) should be able to routed to “HistoryView” when opening the card. 

However, the regular users will be routed to SingupView every time they open the card.

> ℹ️ This happens only with using frontendClient

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/a1a78eba-e669-4369-b63b-4717b7da7be9


juni-notifi (commented):
Is this a bug with frontend client? Or a bug in general?

eric-notifi:
This issue can only be seen when using frontend client.

juni-notifi (approved):

PR #351, opened 2023-07-06
Support deleting in subscribe

juni-notifi:
The internal method already supports a null config, just support it in the batch api


eric-notifi (approved):
👍

PR #353, opened 2023-07-10
Support ZKSYNC

mayuranganesathas:
Aside from adding the ENUMs, how do we ensure that the wallet is supported?

In this case for ZkSync we're using Metamask 

@marukohao

### DIFF
--- a/packages/notifi-core/lib/NotifiClient.ts
+++ b/packages/notifi-core/lib/NotifiClient.ts
@@ -276,7 +276,8 @@ export type SignMessageParams =
         | 'BINANCE'
         | 'INJECTIVE'
         | 'OSMOSIS'
-        | 'OPTIMISM';
+        | 'OPTIMISM'
+        | 'ZKSYNC';
       signMessage: Uint8SignMessageFunction;
     }>
   | Readonly<{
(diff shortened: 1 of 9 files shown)


Nathan-Notifi:
@mayuranganesathas did yhou try signing in to the SDK react card using ZK?

Nathan-Notifi:
@mayuranganesathas i think we need to update the frontend client as well

ype '"INJECTIVE" | "OSMOSIS" | "ZKSYNC" | "APTOS" | "ACALA" | "NEAR" | "SUI"' is not assignable to type 'WalletBlockchain'.
  Type '"ZKSYNC"' is not assignable to type 'WalletBlockchain'.ts(2322)
generated.ts(2955, 3): The expected type comes from property 'walletBlockchain' which is declared here on type 'Exact<{ walletBlockchain: WalletBlockchain; walletPublicKey: string; dappAddress: string; timestamp: any; signature: string; accountId?: InputMaybe<string>; }>'

mayuranganesathas:
Awaiting GraphQL Schema updates to support ZKSYNC

mayuranganesathas:
Seems good to go. Open build issue on main - @juni-notifi w/ alertFrequency param. Should I make it an optional param?

PR #364, opened 2023-08-09
MVP-3062: Throw error explicitly when failing subscribe alert

eric-notifi:
Sometimes, the alert subscription toggle pumps back to "unsubscribe" state after trying to subscribe w/o the reason. 
Throwing error in this case would help user (or dev) to get the hint.

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
+++ b/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
@@ -435,7 +435,11 @@ export const useNotifiSubscribe: ({
           .find((f) => f?.filterType === filterType);
         if (filter === undefined || filter.id === null) {
           await deleteThisAlert();
-          return null;
+          throw new Error(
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -435,7 +435,11 @@ export const useNotifiSubscribe: ({
           .find((f) => f?.filterType === filterType);
         if (filter === undefined || filter.id === null) {
           await deleteThisAlert();
-          return null;
+          throw new Error(
+            `No applicableFilters filter for sources: ${JSON.stringify(
+              sources,
+            )}`,
```
juni-notifi, line 441: do we also want to throw on line 523?

eric-notifi: Yeah, Also added for single source alert (line#523). Thanks for reminding @juni-notifi

juni-notifi (approved):

PR #374, opened 2023-08-28
MVP-3179: support new query GetUnreadNotificationHistoryCount in SDK

eric-notifi:
## Add `GetUnreadNotificationHistoryCount` api available in SDK 
- hooks
- frontend-client 

## Add demo in `notifi-react-example`

### DIFF
--- a/packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
+++ b/packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
@@ -1509,6 +1509,22 @@ const useNotifiClient = (
     [setLoading, setError, service],
   );
 
+  const getUnreadNotificationHistoryCount = useCallback(async () => {
+    try {
+      const result = await service.getUnreadNotificationHistoryCount();
+      return result;
+    } catch (e: unknown) {
+      if (e instanceof Error) {
+        setError(e);
+      } else {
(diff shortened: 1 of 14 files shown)


packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
```diff
@@ -1509,6 +1509,22 @@ const useNotifiClient = (
...
+        setError(e);
+      } else {
+        setError(new NotifiClientError(e));
+      }
+      throw e;
+    } finally {
+      setLoading(false);
+    }
+  }, [setLoading, setError, service]);
+
```
juni-notifi, line 1527: Do you think it would be useful to put the count in `data` as well? Or do you think it's not necessary?

juni-notifi (approved):

packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
eric-notifi: IMO, it is not necessary for now because:
- it is not widely used across views . 
- it changes not often only when the `MarkFusionNotificationHistoryAsRead` is called (at least not now).
It might be good to reconsider to add it in `data` if we want to implement it in `react-card`

PR #376, opened 2023-08-31
fix line break issue in destination section

marukohao:
when the email address is too long, the 'resend link' is broken down into two line
also fixed the 'invalid DOM property' warning from CircleBellIcon
image

before:
Screenshot 2023-08-28 at 3 25 36 PM
after:
Screenshot 2023-08-28 at 3 24 58 PM

### DIFF
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -883,7 +883,10 @@ input::-webkit-inner-spin-button {
   overflow-wrap: anywhere;
 }
 
-.NotifiUserInfoPanel__email {
+.NotifiUserInfoPanel__email,
+.NotifiUserInfoPanel__telegram,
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/components/defaults.css
```diff
@@ -879,6 +883,10 @@ input::-webkit-inner-spin-button {
   overflow-wrap: anywhere;
 }
 
+.NotifiUserInfoPanel__email {
```
juni-notifi, line 886: is this only necessary for email? what about Telegram or SMS?

juni-notifi (approved):

packages/notifi-react-card/lib/components/defaults.css
marukohao: yeah, I think we should add gap on all the destinations

PR #378, opened 2023-09-04
MVP-3180: support new mutation MarkFusionNotificationHistoryAsRead

eric-notifi:
Add mutation methods (frontend-client/ hooks)

Add demo (react-card/ frontend-client)


juni-notifi (approved):
Can you also add a sample for using specific `ids`

eric-notifi:
Added an sample for marking only one (newest) history entry as read. CC. @juni-notifi

PR #380, opened 2023-09-08
MVP-3202: adopt new fusionNotificationHistory (default)

eric-notifi:
- Adopt new `fusionNotificationHistoryEntry` (`includeHidden=false`) endpoint for frontendClient by default (`isUsingFrontendClient=true`)

- Legacy `notificationHistoryEntry` will be used while using `hooks` (`isUsingFrontendClient=false`)

### DIFF
--- a/packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
+++ b/packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
@@ -234,12 +234,12 @@ const concatHistoryNodes = (
   switch (nodes[0]?.__typename) {
     case 'FusionNotificationHistoryEntry':
       if (nodesToConcat[0]?.__typename !== 'FusionNotificationHistoryEntry') {
-        throw new Error('Type mismatch: FusionNotificationHistoryEntry');
+        return nodes;
       }
       return [...nodes, ...nodesToConcat];
     case 'NotificationHistoryEntry':
       if (nodesToConcat[0]?.__typename !== 'NotificationHistoryEntry') {
-        throw new Error('Type mismatch: NotificationHistoryEntry');
+        return nodes;
       }
       return [...nodes, ...nodesToConcat];
     default:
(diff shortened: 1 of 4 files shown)


packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
```diff
@@ -242,8 +227,29 @@ const getAlertDetailsContents = (
...
 };
 
+const concatHistoryNodes = (
+  nodes: NotificationHistoryEntry[],
+  nodesToConcat: NotificationHistoryEntry[],
+) => {
+  switch (nodes[0]?.__typename) {
+    case 'FusionNotificationHistoryEntry':
+      if (nodesToConcat[0]?.__typename !== 'FusionNotificationHistoryEntry') {
+        throw new Error('Type mismatch: FusionNotificationHistoryEntry');
```
juni-notifi, line 237: rather than throwing, can we just return `nodes` here? (i.e. throw away nodesToConcat)

packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
```diff
@@ -242,8 +227,29 @@ const getAlertDetailsContents = (
...
+      }
+      return [...nodes, ...nodesToConcat];
+    case 'NotificationHistoryEntry':
+      if (nodesToConcat[0]?.__typename !== 'NotificationHistoryEntry') {
+        throw new Error('Type mismatch: NotificationHistoryEntry');
```
juni-notifi, line 242: same

packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
eric-notifi: Nice idea👍. modified.

packages/notifi-react-card/lib/utils/NotificationHistoryUtils.tsx
eric-notifi: Modified.

juni-notifi (approved):

PR #381, opened 2023-09-11
MVP-3254: Make use of Nx-build to speed up build time

eric-notifi:
We can use lerna's built-in nx support to cache and speed up local build time.

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/77437648-2fba-48da-b6d0-e119152753d2

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "test": "jest"
   },
   "devDependencies": {
+    "@nrwl/nx-darwin-x64": "^15.9.3",
     "@trivago/prettier-plugin-sort-imports": "^3.3.0",
     "@types/jest": "^27.4.1",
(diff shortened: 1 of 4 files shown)


juni-notifi (approved):

package.json
```diff
@@ -9,6 +9,7 @@
     "test": "jest"
   },
   "devDependencies": {
+    "@nrwl/nx-darwin-x64": "^15.9.3",
```
juni-notifi, line 12: i'm on M1 chip, should i install nx-darwin-arm64?

eric-notifi: Yeah ... seems like we need nx-drawin-arm64.
BTW, my env is also M1

PR #382, opened 2023-09-11
add EventTypeFusionMultiThreshhold

Haoqian-Zheng:
This PR adds a new EventType to the react-card that allows users to set up multiple fusion health check alerts in one row 

Screen Shot 2023-09-12 at 00 29 23

### DIFF
--- a/packages/notifi-frontend-client/lib/client/ensureSource.ts
+++ b/packages/notifi-frontend-client/lib/client/ensureSource.ts
@@ -474,6 +474,9 @@ const ensureSources = async (
     case 'label': {
       throw new Error('Unsupported event type');
     }
+    case 'fusionMultiThreshhold': {
+      throw new Error('Unsupported event type');
+    }
   }
 };
 
@@ -1007,6 +1010,9 @@ export const ensureSourceAndFilters = async (
       };
     }
     case 'label': {
+      throw new Error('Unsupported event type');
+    }
+    case 'fusionMultiThreshhold': {
       throw new Error('Unsupported event type');
     }
     case 'walletBalance': {
--- a/packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
+++ b/packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
@@ -75,6 +75,19 @@ export type TradingPairEventTypeItem = Readonly<{
   tradingPairs: ValueOrRef<ReadonlyArray<string>>;
 }>;
 
+export type FusionMultiThreshholdEventTypeItem = Readonly<{
+  type: 'fusionMultiThreshhold';
+  name: string;
+  tooltipContent?: string;
+  numberType: NumberTypeSelect;
+  subtitle?: string;
+  addThreshholdTitle?: string;
+  fusionEventId: ValueOrRef<string>;
+  sourceAddress: ValueOrRef<string>;
+  maintainSourceGroup?: boolean;
+  alertFrequency?: AlertFrequency;
+}>;
+
 export type PriceChangeDataSource = 'coingecko';
 
 export type PriceChangeEventTypeItem = Readonly<{
@@ -152,6 +165,7 @@ export type EventTypeItem =
   | BroadcastEventTypeItem
   | HealthCheckEventTypeItem
   | TradingPairEventTypeItem
+  | FusionMultiThreshholdEventTypeItem
   | LabelEventTypeItem
   | PriceChangeEventTypeItem
   | CustomTopicTypeItem
--- a/packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
+++ b/packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
@@ -89,6 +89,19 @@ export type TradingPairEventTypeItem = Readonly<{
   tradingPairs: ValueOrRef<ReadonlyArray<string>>;
 }>;
 
+export type FusionMultiThreshholdEventTypeItem = Readonly<{
+  type: 'fusionMultiThreshhold';
+  name: string;
+  tooltipContent?: string;
+  numberType: NumberTypeSelect;
+  subtitle?: string;
+  addThreshholdTitle?: string;
+  fusionEventId: ValueOrRef<string>;
+  sourceAddress: ValueOrRef<string>;
+  maintainSourceGroup?: boolean;
+  alertFrequency?: AlertFrequency;
+}>;
+
 export type WalletBalanceEventTypeItem = Readonly<{
   type: 'walletBalance';
   name: string;
@@ -107,7 +120,7 @@ export type PriceChangeEventTypeItem = Readonly<{
 
 export type USER_INTERFACE_TYPE = 'TOGGLE' | 'HEALTH_CHECK';
 
-export type NumberTypeSelect = 'percentage' | 'integer';
+export type NumberTypeSelect = 'percentage' | 'integer' | 'price';
 
 export type CustomToggleTypeItem = Readonly<{
   filterOptions: FilterOptions;
@@ -151,6 +164,7 @@ export type EventTypeItem =
   | HealthCheckEventTypeItem
   | LabelEventTypeItem
   | TradingPairEventTypeItem
+  | FusionMultiThreshholdEventTypeItem
   | WalletBalanceEventTypeItem
   | PriceChangeEventTypeItem
   | CustomTopicTypeItem
--- a/packages/notifi-react-card/lib/components/defaults.css
+++ b/packages/notifi-react-card/lib/components/defaults.css
@@ -572,6 +572,7 @@ input::-webkit-inner-spin-button {
 .EventTypeCustomToggleRow_label,
 .EventTypeHealthCheckRow__label,
 .EventTypeFusionToggleRow__label,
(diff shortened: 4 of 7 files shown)


packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
```diff
@@ -151,6 +164,7 @@ export type EventTypeItem =
   | HealthCheckEventTypeItem
   | LabelEventTypeItem
   | TradingPairEventTypeItem
+  | FusionMultiThreshholdEventTypeItem
```
eric-notifi, line 167: The same comment as above

packages/notifi-react-card/lib/hooks/SubscriptionCardConfig.ts
```diff
@@ -89,6 +89,19 @@ export type TradingPairEventTypeItem = Readonly<{
...
+  tooltipContent?: string;
+  numberType: NumberTypeSelect;
+  subtitle?: string;
+  addThreshholdTitle?: string;
+  fusionEventId: ValueOrRef<string>;
+  sourceAddress: ValueOrRef<string>;
+  maintainSourceGroup?: boolean;
+  alertFrequency?: AlertFrequency;
+}>;
+
```
eric-notifi, line 104: The same comment as above.

packages/notifi-frontend-client/lib/client/ensureSource.ts
```diff
@@ -474,6 +474,9 @@ const ensureSources = async (
     case 'label': {
       throw new Error('Unsupported event type');
     }
+    case 'fusionMultiThreshhold': {
+      throw new Error('Unsupported event type');
```
eric-notifi, line 478: Oh.. I think there is a typo, maybe `fusionMultiThreshold`?

And because we can reuse the `ensureFusionSource` here, we can just put this case above `case 'fusion'` as below:
```
// ....
    case 'fusionMultiThreshold'
    case 'fusionToggle':
    case 'fusion': {
      const source = await ensureFusionSource(service, eventType, inputs);
      return [source];
    }
```

packages/notifi-frontend-client/lib/client/ensureSource.ts
```diff
@@ -1009,6 +1012,9 @@ export const ensureSourceAndFilters = async (
...
+    }
```
eric-notifi, line 1017: Here inside the ensureSourceAndFilters, we might also want to implement a `getFusionMultiThreshold` method to gather the correct filter and filterOptions.

packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
```diff
@@ -75,6 +75,19 @@ export type TradingPairEventTypeItem = Readonly<{
...
+  tooltipContent?: string;
+  numberType: NumberTypeSelect;
+  subtitle?: string;
+  addThreshholdTitle?: string;
+  fusionEventId: ValueOrRef<string>;
+  sourceAddress: ValueOrRef<string>;
+  maintainSourceGroup?: boolean;
+  alertFrequency?: AlertFrequency;
+}>;
+
```
eric-notifi, line 90: Since it is also under `FusionEventTypeItem`, so I might suggest to re-use `FusionTypeBase` and use the `selectedUIType` to distinguish this one from others. like below
```
export type FusionEventTypeItem =
  | FusionToggleEventTypeItem
  | FusionHealthCheckEventTypeItem
  | FusionMultiThresholdEventTypeItem

 export type FusionMultiThresholdTypeItem = FusionTypeBase &
  Readonly<{
    selectedUIType: 'MULTI_THRESHOLD';
    // ....
  }>;
```

packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
```diff
@@ -152,6 +165,7 @@ export type EventTypeItem =
   | BroadcastEventTypeItem
   | HealthCheckEventTypeItem
   | TradingPairEventTypeItem
+  | FusionMultiThreshholdEventTypeItem
```
eric-notifi, line 168: Here as mentioned above, we will not need to have a new type defined here. We can just reuse common `FusionEventTypeItem`

eric-notifi (changes requested):

packages/notifi-frontend-client/lib/client/ensureSource.ts
Haoqian-Zheng: `fusionMultiThreshold` is not really an alert itself, it is just a wrapper for multiple `fusion` alerts, so I don't think we will ever want to ensureAlert with an actual fusionMultiThreshold eventtypeitem

packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
Haoqian-Zheng: hm based on my comment here https://github.com/notifi-network/notifi-sdk-ts/pull/382#discussion_r1323394058, I think fusionMultiThreshhold should not be a subtype of fusion. 

I can use fusiontypebase with omits to reduce redundancy though

Haoqian-Zheng: i think if fusionMultithreshold is a subtype of fusion, we will run into issues where we will see multiple toggles for the same multithreshold alert when the user signs up for multiple thresholds

packages/notifi-frontend-client/lib/client/ensureSource.ts
eric-notifi: Oh yeah ... it is similar as `TradingPair` that is actually a `DirectPush` source with it's own `filterOption`.
So we might be able to reuse the existing `ensureSource` for fusion alert. 
CC: @juni-notifi: Hope I understand it correctly.

packages/notifi-frontend-client/lib/models/SubscriptionCardConfig.ts
eric-notifi: Not sure what does the `filterOption` of this alert look like. 
I though it might be single fusion source. And then BE uses it's `filterOtpion` to decide what to emit . If it is the case, it would not have multiple toggles (since the each threshold is defined in filterOption)

packages/notifi-frontend-client/lib/client/ensureSource.ts
Haoqian-Zheng: yeah the way i have designed it is that I'm just subscribing a `fusion` source, and never a `fusionMultiThreshold` source, similar to `TradingPair`. In this case every alert is the same as the fusion health check alert so I'm reusing all the fusion filters and sources, thus I have 

`case 'fusionMultiThreshhold': {
      throw new Error('Unsupported event type');
    }`

and also in the subscribeAlert you can see I'm just creating and subscribing a fusion alert https://github.com/notifi-network/notifi-sdk-ts/pull/382/files#diff-ec97d4ca3cd1d1913fc81a7184a0520b17c17d4c913688eadc505b43231702e7R315

eric-notifi: Oh ... I see, we convert `FusionMultiThresholdEventType` to `FusionEventTypeItem`  (HEALTH_CHECK). Got it.

That case, do we have to define a new type (`FusionMultiThresholdEventType`). or can we reuse the fusion health type simply.

Haoqian-Zheng: I think we need the `FusionMultiThresholdEventType` to render the correct component in the card, otherwise it would render the fusion health check row component

eric-notifi: The last question is that why cannot we reuse the `FusionEventTypeItem` with `FusionTypeBase`.
How do you think using `FusionEventTypeItem` with `FusionTypeBase` and define the `selectedUIType=MULTI_THRESHHOLD`. Then instead of throwing error in `ensureSourceAndFilters`, we can just simply reuse `ensureFusionSource`.
And just need to change `getFusionSouceFilter` as below.

````ts
const getFusionSourceFilter = (
  //...
  else if (eventType.selectedUIType === 'HEALTH_CHECK' || eventType.selectedUIType === 'MULTI_THRESHHOLD') {
    // ....
  }
```

Haoqian-Zheng: I dont think we would ever want to `ensureSourceAndFilters` with a multithreshold eventtype, since we only want one (so only one row renders in the card) so I would definitely keep throwing the error in `ensureSourceAndFilters`. 

However I think we can reuse `FusionTypeBase` at the type declaration to reduce redundancy and use `selectedUIType` to make sure the right component renders, let me look into that

Haoqian-Zheng: added a change to use `FusionTypeBase`, for the getFusionSourceFilter I don't think we want to ever call ensureAlert with a fusionMulti event type so I would not change it

Haoqian-Zheng:
@eric-notifi this PR doesn't get the latest commits on the branch for some reason, moving this to https://github.com/notifi-network/notifi-sdk-ts/pull/383

PR #383, opened 2023-09-14
add EventTypeFusionMultiThreshhold

Haoqian-Zheng:
This PR is a dupe of https://github.com/notifi-network/notifi-sdk-ts/pull/382 which did not get the latest commits in the branch for unknown reasons. 

This PR adds a new EventType to the react-card that allows users to set up multiple fusion health check alerts in one row

https://user-images.githubusercontent.com/8414510/267162445-11aee9a7-5734-48ad-afc8-66e98930631b.png

### DIFF
--- a/packages/notifi-frontend-client/lib/client/ensureSource.ts
+++ b/packages/notifi-frontend-client/lib/client/ensureSource.ts
@@ -474,9 +474,6 @@ const ensureSources = async (
     case 'label': {
       throw new Error('Unsupported event type');
     }
-    case 'fusionMultiThreshold': {
-      throw new Error('Unsupported event type');
-    }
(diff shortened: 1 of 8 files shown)


eric-notifi (approved):
LGTM !! 🙂
Just a small nit that I think we might need to add `|| eventType.selectedUIType === "MULTI_THRESHOLD` [here](https://github.com/notifi-network/notifi-sdk-ts/blob/aef86fc5650105dc44e2e64f06b85c346a072239/packages/notifi-frontend-client/lib/client/ensureSource.ts#L685) in order to return the correct `filterOptions`. Otherwise, the `filterOptions` will always be empty object

PR #392, opened 2023-10-04
MVP-3254-patch: replace devDependency '@nrwl/nx-darwin-x64' with more general one 'nx'

eric-notifi:
According to the [official doc](https://nx.dev/recipes/adopting-nx/lerna-and-nx#use-nx-for-task-scheduling-without-changing-the-lerna-setup), we should just install "nx" as dev dependency instead of the `@nrwl/nx-darwin-x64` which is specifically for macOS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@nrwl/nx-darwin-x64": "^15.9.3",
+    "nx": "^16.9.1",
     "@trivago/prettier-plugin-sort-imports": "^3.3.0",
     "@types/jest": "^27.4.1",
     "@types/node-fetch": "^2.6.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,6 @@
         "packages/notifi-solana-hw-login"
       ],
       "devDependencies": {
-        "@nrwl/nx-darwin-x64": "^15.9.3",
         "@trivago/prettier-plugin-sort-imports": "^3.3.0",
         "@types/jest": "^27.4.1",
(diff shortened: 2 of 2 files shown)


juni-notifi (approved):

package.json
```diff
@@ -9,7 +9,7 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@nrwl/nx-darwin-x64": "^15.9.3",
+    "nx": "^16.9.1",
```
juni-notifi, line 12: oh is it okay that we're going up to vs 16 from 15?

juni-notifi (changes requested):

package.json
eric-notifi: Yeah ... I think it is ok. 
The newest version of @nrwl/nx-darwin-x64 is 15.9.7 and`nx` also depends on `15.9.7` version of @nrwl/nx-darwin-x64. Meaning the @nrwl/nx-darwin-x64 15.9.7 will be also installed when installing `nx`
I tested the command that we are using often (build/ publish). They all work fine.
cc: @juni-notifi

eric-notifi: I think @nrwl/nx-darwin-x64 version is not related to nx main version. It is difficult to fig out their dependency chain. 
But I realized when I install `nx` and `@nrwl/nx-darwin-x64` v15 and `@nx/nx-darwin-x64` v16 are both installed. 
And surprisingly they are both sharing the same content (`nx.darwin-x64.node`).
I tried all our often used commands (build/ publish). They both work fine with `nx` v16.

juni-notifi (approved):

PR #395, opened 2023-10-11
MVP-3445: Fix Solana 'use hardwallet' not showing when destination is not required

eric-notifi:
In the react-card, when “Require Destinations” is turned off the signup page does not render the input fields including the “use hardware wallet” toggle for Solana with the hw wallet package. However the user needs to sign a message to login so will end up in a locked state where they cannot register at all. This will prevent any Solana users using a hw wallet from signing up to any dapp notifications. We should render the “use hardware wallet” toggle even if require destination is turned off.


Haoqian-Zheng (approved):

juni-notifi:
merging to cut a new release

PR #398, opened 2023-10-12
fix above for fusionmultithreshold signup (synthetix cratio)

Haoqian-Zheng:


packages/notifi-react-card/lib/components/subscription/EventTypeFusionMultiThresholdRow.tsx
```diff
@@ -306,6 +306,7 @@ export const FusionMultiThresholdSettingsRow: React.FC<
         inputs: Record<string, unknown>;
       }>,
       threshold: number,
+      above: boolean,
```
juni-notifi, line 309: oh, rather than passing as a method parameter, we should add `above` to the dependencies array on line 341

I think that's preferable to shadowing the variable in the containing scope.

While you're at it, can you please add `alertConfiguration` to the dependencies array too?

We really need to look into adding lint errors for missing dependencies

Haoqian-Zheng: messed up the branches, moving this to https://github.com/notifi-network/notifi-sdk-ts/pull/399

Haoqian-Zheng:
moving this to https://github.com/notifi-network/notifi-sdk-ts/pull/399

PR #402, opened 2023-10-17
Use discord server invite link from target

kai-notifi:


packages/notifi-graphql/lib/gql/fragments/DiscordTargetFragment.gql.ts
```diff
@@ -10,5 +10,6 @@ export const DiscordTargetFragment = gql`
     name
     userStatus
     verificationLink
+    discordServerInviteLink
```
juni-notifi, line 13: we need to add it to `notifi-axios-adapter/lib/fragments/discordTargetFragment.ts` as well

juni-notifi (approved):

PR #404, opened 2023-10-18
fix can not toggle off discord target

marukohao:
when there is a discord target exists(discord id is exists), we always reset the useDiscord to true, even if the user has been toggles off the discord from edit view. 
`contactInfo?.discord?.active && useDiscord` is the logic for showing discord target, useDiscord === true should only be set when user switches on from the toggle button


eric-notifi:
@marukohao 
When the discord is toggled off, the discordTarget in TargetGroup should be empty array. So it should not get into `if (discordId) {` block and the useDiscord should not be set to turn again. Not sure what issue you are encountering. 
I just had a test. It is working normally.

juni-notifi:
I think we can close this, right?

PR #405, opened 2023-10-19
Fix discord target in NotifiSubscribeButton

juni-notifi:
The button had unnecessary logic for resolving the discord target.

This is probably stemming from the misnamed "discordId" param

Simply rely on ensureTargetGroup to do the right thing

### DIFF
--- a/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
+++ b/packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
@@ -713,13 +713,6 @@ export const useNotifiSubscribe: ({
 
     let finalPhoneNumber = undefined;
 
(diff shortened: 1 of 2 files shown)


packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts
```diff
@@ -713,13 +713,6 @@ export const useNotifiSubscribe: ({
 
     let finalPhoneNumber = undefined;
 
-    let finalDiscordId = undefined;
-
-    if (useDiscord === true) {
-      finalDiscordId =
```
juni-notifi, line 719: finalDiscordId was unused

kai-notifi (approved):

PR #409, opened 2023-10-20
MVP-3439: display VerifyBanner while destination is not verified 

eric-notifi:
## New FTU flow #3: Display VerifyBanner while destination is not verified 
See the demo video below

Screenshot 2023-10-20 at 14 36 50

## IMPORTANT NOTE
This is only one out of total flows of New FTU. We will merge together all done (MVP-3437/ MVP-3438/ MVP-3439).


packages/notifi-react-card/lib/assets/CheckIcon.tsx
```diff
@@ -1,11 +1,13 @@
 import React from 'react';
 
-export type Props = React.SVGProps<SVGSVGElement>;
+export type Props = Readonly<{
+  className?: string;
+}>;
```
juni-notifi, line 5: hmmm i think in general it's nice to be able to override other props, like width and height

juni-notifi (approved):

packages/notifi-react-card/lib/assets/CheckIcon.tsx
eric-notifi: Modified: Allow to pass in svg props into CheckIcon
```
export type Props = Readonly<
  {
    className?: string;
  } & React.SVGProps<SVGSVGElement>
>;
```

PR #410, opened 2023-10-26
fix thresholdDirection

Haoqian-Zheng:


packages/notifi-react-card/lib/components/subscription/EventTypeFusionHealthCheckRow.tsx
```diff
@@ -176,7 +176,7 @@ export const EventTypeFusionHealthCheckRow: React.FC<
           ratioNumber;
         alertDetail.inputs[
           `${alertDetail.eventType.name}__healthThresholdDirection`
-        ] = thresholdDirection;
+        ] = config.checkRatios[0].type ?? thresholdDirection;
```
juni-notifi, line 179: to be safe, can we do `config.checkRatios[0]?.type`

juni-notifi (approved):

PR #414, opened 2023-10-30
add support for displayname to preview card

Haoqian-Zheng:


packages/notifi-react-card/lib/components/subscription/subscription-card-views/AlertListPreview.tsx
```diff
@@ -47,7 +47,7 @@ export const AlertListPreview: React.FC<AlertListPreviewProps> = ({
           )}
           key={eventType.name}
         >
-          {eventType.name}
+          {eventType.type === 'fusion' ? (eventType.displayNameOverride ?? eventType.name) : eventType.name}
```
juni-notifi, line 50: don't we also have fusionHealthCheck

juni-notifi (approved):

packages/notifi-react-card/lib/components/subscription/subscription-card-views/AlertListPreview.tsx
Haoqian-Zheng: that is just a sub type of `fusion` i think

PR #418, opened 2023-11-08
MVP-3442: CI for cypress E2E test

eric-notifi:
As title, Add github action for cypress e2e CI

### DIFF
--- a/packages/notifi-react-example/package.json
+++ b/packages/notifi-react-example/package.json
@@ -43,7 +43,6 @@
   },
   "scripts": {
     "start": "BROWSER=none react-app-rewired start",
-    "build": "react-app-rewired build",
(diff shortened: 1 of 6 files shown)


packages/notifi-react-example/package.json
```diff
@@ -43,10 +43,10 @@
   },
   "scripts": {
     "start": "BROWSER=none react-app-rewired start",
-    "build": "react-app-rewired build",
```
eric-notifi, line 46: We do not need to build notifi-react-example. for the following performance benefits
- local build speed and potential error (warning)
- CI (Github action) faster and avoid potential error and warning.

juni-notifi (approved):

PR #424, opened 2023-11-30
MVP-3735: Add invalid range for fusionHealth

eric-notifi:
Add new `validInputRange` in config type so that we can constraint the upper and lower bounds for health check input value.

No need

MVP-3735

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/EventTypeFusionHealthCheckRow.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/EventTypeFusionHealthCheckRow.tsx
@@ -2,6 +2,7 @@ import {
   EventTypeItem,
   FusionHealthCheckEventTypeItem,
   ThresholdDirection,
+  ValidInputRange,
 } from '@notifi-network/notifi-frontend-client';
 import clsx from 'clsx';
 import React, {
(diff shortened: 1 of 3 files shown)


packages/notifi-react-card/lib/components/subscription/EventTypeFusionHealthCheckRow.tsx
```diff
@@ -99,6 +100,9 @@ export const EventTypeFusionHealthCheckRow: React.FC<
   const UNABLE_TO_SUBSCRIBE = 'Unable to subscribe, please try again';
   const UNABLE_TO_UNSUBSCRIBE = 'Unable to unsubscribe, please try again';
   const INVALID_NUMBER = 'Please enter a valid number';
+  const INVALID_RANGE = (validInputRange: ValidInputRange) => {
+    return `Please enter a valid number (${validInputRange.min} - ${validInputRange.max})`;
```
juni-notifi, line 104: is this copy okay? maybe something like "Please enter a number between {min} and {max}"

packages/notifi-react-card/lib/components/subscription/EventTypeFusionHealthCheckRow.tsx
```diff
@@ -219,42 +246,41 @@ export const EventTypeFusionHealthCheckRow: React.FC<
...
-      case 'price':
-        ratioNumber = getParsedPrice(customInputRef.current.value);
-        break;
-      default: // 'integer'
-        ratioNumber = parseFloat(customInputRef.current.value);
-    }
-
-    if (ratioNumber && ratioNumber >= 0 && customValue) {
-      subscribeAlert({ eventType: config, inputs }, ratioNumber)
+    if (inputNumber && inputNumber >= 0 && customValue) {
```
juni-notifi, line 282: why do we need to check `inputNumber >= 0`?

eric-notifi: Right, we should not restrict users not to subscribe 0 since we now able to input valid number range.

juni-notifi (approved):

PR #425, opened 2023-11-30
Add html to generic event details

kai-notifi:
Add html content for generic message html. This is to support html for the new fusion broadcast that will be generic event details instead of broadcast event details.


eric-notifi (commented):
Looks good but seems ` is not yet on `prod`. This causes the build fail, shall we merge it after being available.

kai-notifi:
> Looks good but seems `is not yet on`prod`. This causes the build fail, shall we merge it after being available.

Yep!

PR #431, opened 2023-12-15
MVP-3643: new cardView historyDetail

eric-notifi:

As the the following situations of history cardView , it becomes more and more difficult to follow and maintain.

includes not only history but also history detail view.

too much conditional logic between history and detail view

To get around this situation, we need to refactor history section in SubscriptionCardV1 by separating to hisotry & historyDetail.

PASS

MVP-3643

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
@@ -1,6 +1,7 @@
 import { CardConfigItemV1 } from '@notifi-network/notifi-frontend-client';
 import { Types } from '@notifi-network/notifi-graphql';
 import clsx from 'clsx';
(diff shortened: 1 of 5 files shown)


packages/notifi-react-card/lib/components/subscription/subscription-card-views/HistoryCardView.tsx
```diff
@@ -23,49 +27,72 @@ import {
...
+    notificationImage?: string;
+    notificationList?: string;
+    emptyAlertsBellIcon?: string;
+    historyContainer?: string;
+    virtuoso?: string;
+    AlertCard?: AlertNotificationViewProps['classNames'];
+    /**
+     * @deprecated Use `loadingSpinner` instead
+     */
+    LoadingStateCard?: LoadingStateCardProps['classNames'];
```
juni-notifi, line 77: let's remove this! people that upgrade SDK will need to change their styling anyway

juni-notifi (approved):

PR #432, opened 2023-12-15
Base Support

ghost:
Added string literals for base support: note, need to regenerate the gql client before merge


eric-notifi (changes requested):
1. We might want to also add `BASE` into this file `packages/notifi-core/lib/models/WalletBlockchain.ts`

2. I think we might also want to implement `signMessage` method for the new chain  in `packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts`

eric-notifi (approved):
LGTM

PR #435, opened 2023-12-20
MVP-3829: Optimize NotifiClientContext (potential racing issue)

eric-notifi:
Calling object's Async method in `useMemo` before returning it could cause [potential racing issue](https://github.com/notifi-network/notifi-sdk-ts/blob/a38ed62e3d6fb09b0a6b1f3e1a6dda8c6ccb860a/packages/notifi-react-card/lib/context/NotifiClientContext.tsx#L54).  
We should use `useEffect` instead 

No need. (Test PASSED)
MVP-3829

### DIFF
--- a/packages/notifi-react-card/lib/context/NotifiClientContext.tsx
+++ b/packages/notifi-react-card/lib/context/NotifiClientContext.tsx
@@ -68,21 +68,19 @@ export const NotifiClientContextProvider: React.FC<NotifiParams> = ({
     frontendClient.initialize();
   }, [frontendClient]);
 
+  if (!isClientInitialized) return null;
+
   return (
-    <>
-      {isClientInitialized ? (
-        <NotifiClientContext.Provider
-          value={{
-            client,
-            params,
-            isUsingFrontendClient: params.isUsingFrontendClient ?? true,
-            frontendClient,
-          }}
-        >
-          {children}
-        </NotifiClientContext.Provider>
-      ) : null}
-    </>
+    <NotifiClientContext.Provider
+      value={{
+        client,
+        params,
+        isUsingFrontendClient: params.isUsingFrontendClient ?? true,
+        frontendClient,
+      }}
+    >
+      {children}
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/context/NotifiClientContext.tsx
```diff
@@ -47,19 +47,21 @@ export const NotifiClientContextProvider: React.FC<NotifiParams> = ({
...
+    return updatedFrontendClient;
   }, [
     params.dappAddress,
     params.env,
     params.walletBlockchain,
     params.walletPublicKey,
   ]);
 
+  useEffect(() => {
+    frontendClient.initialize();
```
juni-notifi, line 62: race condition still exists for consumers of frontendClient, if consumer uses frontendClient before this initialize completes

juni-notifi (commented):
Hmm I don't know if this actually fixes any race conditions related to `initialize` being complete or not -- let's re-design the Context not to render any children unless the frontendClient has completed initializing

packages/notifi-react-card/lib/context/NotifiClientContext.tsx
eric-notifi: @juni-notifi Great idea, Thanks! Modified it accordingly.

packages/notifi-react-card/lib/context/NotifiClientContext.tsx
```diff
@@ -47,30 +47,42 @@ export const NotifiClientContextProvider: React.FC<NotifiParams> = ({
...
+            client,
+            params,
+            isUsingFrontendClient: params.isUsingFrontendClient ?? true,
+            frontendClient,
+          }}
+        >
+          {children}
+        </NotifiClientContext.Provider>
+      ) : null}
+    </>
```
juni-notifi, line 85: rather than the fragment, i think you can just return early

juni-notifi (approved):

PR #451, opened 2024-01-17
Use the Notifi Dataplane API to call publishFusionMessage

hans-notifi:

The API call `publishFusionMessage` has been migrated away from Notifi's GraphQL API (at `https://api.notifi.network/gql`) to the Dataplane REST API (at `https://dpapi.prd.notifi.network/`). This adapts the SDK to match.


packages/notifi-graphql/lib/NotifiService.ts
```diff
@@ -1,76 +1,79 @@
 import { GraphQLClient } from 'graphql-request';
+import { NotifiDataplaneClient } from '@notifi-network/notifi-dataplane';
 import { v4 as uuid } from 'uuid';
 
 import { version } from '../package.json';
 import * as Generated from './gql/generated';
 import { getSdk } from './gql/generated';
 import type * as Operations from './operations';
+import { FusionMessage } from 'notifi-dataplane/lib/types/FusionMessage';
```
nimesh-notifi, line 9: Do these need to be here?

PR #455, opened 2024-01-25
MVP-3944: Make fusionMessage targetPayload strong type

eric-notifi:
As title: Make fusionMessage targetPayload strong type

No Need (TEST PASSED)

MVP-3944

**DO NOT MERGE**: BE still working on specificWallets type

### DIFF
--- a/packages/notifi-dataplane/lib/types/FusionMessage.ts
+++ b/packages/notifi-dataplane/lib/types/FusionMessage.ts
@@ -1,6 +1,10 @@
 import { Types } from 'notifi-graphql/dist';
 
-export interface FusionMessage {
+export type FusionMessage<T> = T extends infer CommunityFusionMessage
+  ? CommunityFusionMessage
+  : object;
+
+export interface CommunityFusionMessage {
   eventTypeId: string;
   variablesJson: VariablesJsonPayload;
   specificWallets?: ReadonlyArray<
--- a/packages/notifi-dataplane/lib/types/FusionMessage.ts
+++ b/packages/notifi-dataplane/lib/types/FusionMessage.ts
@@ -2,7 +2,7 @@ import { Types } from 'notifi-graphql/dist';
 
 export interface FusionMessage {
   eventTypeId: string;
-  variablesJson: any;
+  variablesJson: VariablesJsonPayload;
   specificWallets?: ReadonlyArray<
     Readonly<{
       walletPublicKey: string;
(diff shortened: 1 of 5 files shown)


kai-notifi (approved):

packages/notifi-dataplane/lib/types/FusionMessage.ts
```diff
@@ -24,3 +24,20 @@ export interface FusionMessage {
...
+/**
+ * @param message - only plain text is supported.
+ * @param message__markdown - markdown string here will be converted to html.
+ */
+export type TargetVariables = {
+  subject?: string;
+  message?: string;
+  message__markdown?: string;
+  [key: string]: string | undefined;
+};
```
hans-notifi, line 43: This type shape is required for Community Manager posts, but not for OffChain API events. Offchain API events can be any JSON as long as the template is prepared for it.

hans-notifi (changes requested):

packages/notifi-dataplane/lib/types/FusionMessage.ts
eric-notifi: @hans-notifi Right ... So we make all field optional in `TargetVariables`. And also allow to put in whatever key-value in it. Or is there any other concern?

hans-notifi: As discussed in Slack, the platform discriminators are not needed for OffChain API events either. You can literally put in any valid JSON and it will work as long as you have a matching template - and for some message types, we would want to do exactly that in order to avoid duplicating data for each platform we want to send on.

packages/notifi-dataplane/lib/types/FusionMessage.ts
```diff
@@ -2,11 +2,28 @@ import { Types } from 'notifi-graphql/dist';
...
     Readonly<{
       walletPublicKey: string;
       walletBlockchain: Types.WalletBlockchain;
     }>
   >;
 }
+
+export type OptionalTargetType = 'Sms' | 'Telegram' | 'Discord' | 'Email';
+
+export type VariablesJsonPayload = { Platform: TargetVariables } & {
```
hans-notifi, line 16: Nit: As discussed, I think it's good exposing it like this. I'd suggest renaming it `CommunityManagerVariablesJsonPayload` in order to specifically indicate that this is for Community Manager posts within the code (rather than create any impression that all messages should look like this).

packages/notifi-node-sample/README.md
```diff
@@ -149,6 +149,11 @@ If the request is successful, we can get a response body like this:
 }
 ```
 
+> NOTE:
+> The `variablesJson` could vary depending on situations.
+> For Community Manager posts: Follow the type `VariablesJsonPayload`
+> For OffChain parser, it could be any shape of json object associated with the template
```
hans-notifi, line 155: Nit: I'd suggest the following copy here:

```
The `variablesJson` parameter is the set of variables that will be used when rendering your templates.
If you have a variable `fromAddress` in there, for example, you can display it in the template with the
expression `{{ eventData.fromAddress }}`.
The type `CommunityManagerVariablesJsonPayload` defines what this object should look like for
the templates that Community Manager posts use.
```

hans-notifi (approved):
Looks great! Just some small nits for variable naming and message copy.

PR #456, opened 2024-01-25
MVP-3946: Refactor Dataplane publishFusionMessage

eric-notifi:
- IMO, Data plane api supposes to be parallel with GQL service.  So ...

    - To avoid confusion, we might not want to include dataplane method (publishFusionMessage) into notifi-graphql package

    - notifi-graphql does not need to depend on notifi-dataplane anymore, so there will no longer be circular reference issue in notifi-dataplane pakcage

- Since we now have notifi-dataplane package, we want to pass in two services (gql & dataplane) to create a NotifiClient in notifi-node package

- Remove the duplicated module in notifi-node

NO NEED (PASSED)

MVP-3946


hans-notifi (approved):
Looks good to me. I originally didn't do it this way because I wanted to avoid breaking changes, but since it looks like we're fine with them, I think this is better.

PR #457, opened 2024-01-25
Change specificWallets to match the shape of the dataplane API

hans-notifi:
The `FusionIngest` endpoint in the Dataplane API takes specific wallet targets as a JSON object with the wallet addresses as keys and the wallet blockchain names as values. This changes that API to match.

No need


hans-notifi:
Closing this in favor of an internal PR that changes the API shape of the service instead.

PR #461, opened 2024-01-30
OpenDefi/MVP-3954: Build up infra for new page `notifi-dapp-example`

eric-notifi:
As title:  Build up infra for new page `notifi-dapp-example`

No need

MVP-3954


packages/notifi-dapp-example/src/app/globals.css
```diff
@@ -0,0 +1,72 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
+
+@layer base {
+  :root {
+    /* Notifi global color */
```
eric-notifi, line 7: We still use css global variables to manage color.
This way will be able to have dark/mode switch if needed.

eric-notifi (commented):
We have mix css global color variable to have more flexibility. (see detailed info inline)

PR #472, opened 2024-02-13
OpenDefi/MVP-3961: add Dashboard page

eric-notifi:
  Implement dashboard and relevant components

No need
MVP-3961


packages/notifi-dapp-example/src/app/notifi/dashboard/page.tsx
```diff
@@ -1,58 +1,68 @@
...
+import { WalletAccount } from '@cosmos-kit/core';
+import { useWalletClient } from '@cosmos-kit/react';
+import { useEffect, useState } from 'react';
+
+export type CardView =
+  | 'history'
+  | 'historyDetail'
+  | 'destination'
+  | 'alertSubscription'
+  | 'editDestination';
```
marukohao, line 17: Do we need the 'editDestination' cardView, it can show as the modal in 'destination' view

marukohao (approved):
LGTM

packages/notifi-dapp-example/src/app/notifi/dashboard/page.tsx
eric-notifi: Well, I think we might not need to have that if it will use popup modal, so I removed it. 
With that being said, feel free add any needed view if needed

PR #475, opened 2024-02-14
Allow variables in FusionMessage to have any shape

hans-notifi:
- Events attached to Community Manager have a specific message shape that segregates data by platform, while events of type "Sent from my server" have no such shape restriction.
- Hence, change the type of `variablesJson` in FusionMessage to `object` and remove `CommunityManagerJsonPayload`

### DIFF
--- a/packages/notifi-dataplane/lib/types/FusionMessage.ts
+++ b/packages/notifi-dataplane/lib/types/FusionMessage.ts
@@ -1,9 +1,8 @@
 import { Types } from 'notifi-graphql/dist';
 
-// If payload type undefined, fallback to default CommunityManagerJsonPayload
-export interface FusionMessage<T extends object = CommunityManagerJsonPayload> {
+export interface FusionMessage {
   eventTypeId: string;
-  variablesJson: VariablesJsonPayload<T>;
+  variablesJson: object;
   specificWallets?: ReadonlyArray<
     Readonly<{
       walletPublicKey: string;
@@ -11,23 +10,3 @@ export interface FusionMessage<T extends object = CommunityManagerJsonPayload> {
     }>
   >;
 }
-
-export type VariablesJsonPayload<T extends object> =
-  T extends CommunityManagerJsonPayload ? CommunityManagerJsonPayload : object;
-
-export type CommunityManagerJsonPayload = { Platform: TargetVariables } & {
-  [key in OptionalTargetType]?: TargetVariables;
-};
-
-export type OptionalTargetType = 'Sms' | 'Telegram' | 'Discord' | 'Email';
-
-/**
- * @param message - only plain text is supported.
- * @param message__markdown - markdown string here will be converted to html.
- */
-export type TargetVariables = {
-  subject?: string;
-  message?: string;
-  message__markdown?: string;
-  [key: string]: string | undefined;
-};
--- a/packages/notifi-dataplane/lib/types/CommunityManagerVariablesJsonPayload.ts
+++ b/packages/notifi-dataplane/lib/types/CommunityManagerVariablesJsonPayload.ts
@@ -1,0 +1,41 @@
+/**
+ * The variable shape that is used for publishing Fusion messages
+ * to Community Manager topics in `publishFusionMessage()`, used as
+ * the `variablesJson` field in `FusionMessage`
+ */
+export type CommunityManagerVariablesJsonPayload = {
+  Platform: CommunityManagerTargetVariables;
+} & {
+  [key in CommunityManagerOptionalTargetType]?: CommunityManagerTargetVariables;
+};
+
+/**
+ * An additional target type to send the Community Manager message to
+ */
+export type CommunityManagerOptionalTargetType =
+  | 'Sms'
+  | 'Telegram'
+  | 'Discord'
+  | 'Email';
+
+/**
+ * The set of variables provided for each target in Community manager messages
+ */
+export type CommunityManagerTargetVariables = {
+  /**
+   * The subject line for the message, if applicable
+   */
+  subject?: string;
+  /**
+   * The message, in either HTML or plain text
+   */
+  message?: string;
+  /**
+   * The message in Markdown format. If provided, will override `message`
+   */
+  message__markdown?: string;
+  /**
+   * Additional data to be provided with the message
+   */
+  [key: string]: string | undefined;
+};
--- a/packages/notifi-node-sample/README.md
+++ b/packages/notifi-node-sample/README.md
@@ -151,8 +151,7 @@ If the request is successful, we can get a response body like this:
 
 > NOTE:
 > The `variablesJson` parameter is the set of variables that will be used when rendering your templates.
-> If you have a variable `fromAddress` in there, for example, you can display it in the template with the expression `{{ eventData.fromAddress }}`
-> The type `CommunityManagerJsonPayload` defines what this object should look like for the templates that Community Manager posts use.
+> If you have a variable `fromAddress`, for example, you can display it in the template with the expression `{{ eventData.fromAddress }}`
 
 If the request is successful, we can get a response body like this:
 


hans-notifi:
@eric-notifi I recall that we set up this type restriction in https://github.com/notifi-network/notifi-sdk-ts/pull/455. However, I had a discussion with the product team and found out that we are not actually intending to use the Node SDK to push to Community Manager topics at all, and are not exposing the Event Type ID in the Admin Portal. Because of that, we don't need to specify any specific shape on the variables being passed into the template.

eric-notifi:
@hans-notifi, appreciate the information. IMOn, retaining that part could be beneficial, as it helps developers understand the potential shape of the CM post. For instance, the Admin Portal development team might find it useful for integration into their codebase. CC: @marukohao @kai-notifi 
If there are no other concerns, shall we consider leaving it as is?

hans-notifi:
> @hans-notifi, appreciate the information. IMO, retaining that part could be beneficial, as it helps developers understand the potential shape of the CM post. For instance, the Admin Portal development team might find it useful for integration into their codebase. CC: @marukohao @kai-notifi 
>
> If there are no other concerns, shall we consider leaving it as is?

There's actually a problem as things currently stand in the SDK: because we're not piping the generic parameter for `FusionMessage` all the way through from the `NotifiClient.publishFusionMessage` call, current users are _stuck_ making Community Manager-formatted posts, even though the SDK _will not_ be used to make them. I double-checked about whether we would need to work with that generic parameter, but since I got confirmation that we would not be using the SDK to make Community Manager posts, I just removed the type bounds instead.

I might have missed something about how get the generic parameters to work though. Do you have some test code that successfully calls `publishFusionMessage` with a different variable type, such as `{ x: number, y: number }`?

I think the CM shape is valuable to document, though, even if we don't keep it in here. We should probably find a good spot in our [Github Pages site](https://github.com/notifi-network/notifi-network.github.io) for that either way.

packages/notifi-dataplane/lib/types/FusionMessage.ts
```diff
@@ -1,33 +1,12 @@
...
   eventTypeId: string;
-  variablesJson: VariablesJsonPayload<T>;
+  variablesJson: object;
   specificWallets?: ReadonlyArray<
     Readonly<{
       walletPublicKey: string;
       walletBlockchain: Types.WalletBlockchain;
     }>
   >;
 }
```
eric-notifi, line 12: @hans-notifi , Just realized you are right, there is a mistaking forcing the Generic in `FusionMessage` as `CommunityManagerJsonPayload`. 
We will be able to fix it by: 

```suggestion
export interface FusionMessage<T extends object = object> {
  eventTypeId: string;
  variablesJson: VariablesJsonPayload<
    T extends CommunityManagerJsonPayload ? CommunityManagerJsonPayload : object
  >;
  specificWallets?: ReadonlyArray<
    Readonly<{
      walletPublicKey: string;
      walletBlockchain: Types.WalletBlockchain;
    }>
  >;
}
```
Then it will allow putting any shape of object ex.

packages/notifi-dataplane/lib/types/FusionMessage.ts
```diff
@@ -1,33 +1,12 @@
 import { Types } from 'notifi-graphql/dist';
 
-// If payload type undefined, fallback to default CommunityManagerJsonPayload
-export interface FusionMessage<T extends object = CommunityManagerJsonPayload> {
+export interface FusionMessage {
```
eric-notifi, line 3:
```suggestion
export interface FusionMessage<T extends object = object> {
```

packages/notifi-dataplane/lib/types/FusionMessage.ts
```diff
@@ -1,33 +1,12 @@
...
   eventTypeId: string;
-  variablesJson: VariablesJsonPayload<T>;
+  variablesJson: object;
```
eric-notifi, line 5:
```suggestion
  variablesJson: VariablesJsonPayload<
    T extends CommunityManagerJsonPayload ? CommunityManagerJsonPayload : object
  >;
```

packages/notifi-dataplane/lib/types/FusionMessage.ts
eric-notifi: Here is an example after change:
```ts
const vars: Readonly<FusionMessage<{ x: number; y: number }>[]> = [
    {
      eventTypeId: 'eventTypeId',
      variablesJson: { x: 1, y: 2 },
    },
  ];


  return client
    .publishFusionMessage(jwt, vars)
```

Screenshot 2024-02-16 at 10 07 08

eric-notifi:
Here is an example after change:
```ts
const vars: Readonly<FusionMessage<{ x: number; y: number }>[]> = [
    {
      eventTypeId: 'eventTypeId',
      variablesJson: { x: 1, y: 2 },
    },
  ];


  return client
    .publishFusionMessage(jwt, vars)
```

Screenshot 2024-02-16 at 10 07 08

And yeah ... I agree that it is important to put it somewhere in our doc in parallel, it will make thing more clear.

hans-notifi:
> Here is an example after change:
> 
> ```ts
> const vars: Readonly<FusionMessage<{ x: number; y: number }>[]> = [
>     {
>       eventTypeId: 'eventTypeId',
>       variablesJson: { x: 1, y: 2 },
>     },
>   ];
> 
> 
>   return client
>     .publishFusionMessage(jwt, vars)
> ```
> 
> Screenshot 2024-02-16 at 10 07 08

I don't think this actually helps us practically. You're putting a type tag on FusionMessage, but it's not actually helping us typecheck anything - TypeScript is only able to give us that information about the shape of the variables because we declared them directly in our source file. I feel that the place that would be useful for inserting type parameters for FusionMessage, if we choose to do that, is if they are on `publishFusionMessage` itself, as in:

```
const vars: Readonly<FusionMessage<{ x: number; y: number }>[]> = [
    {
      eventTypeId: 'eventTypeId',
      variablesJson: { x: 1, y: 2 },
    },
  ];

  // compiles
  const result1 = await client.publishFusionMessage<{ x: number, y: number }>(jwt, vars)
  // does _not_ compile, since z is missing and y is mismatched
  const result2 = await client.publishFusionMessage<{ x: number, y: string, z: number }>(jwt, vars)
```

I couldn't figure out how to get that syntax to work, though, since `publishFusionMessage` is a lambda property on `NotifiClient`.

eric-notifi:
@hans-notifi  The reason you are not able to input `{x: 1, z: 2}` is that you explicitly define the generic type in `FusionMessage`. If we do not define the generic type, it will fall back to `object` type and allow whatever shape we want as long as it is an `object`. example:
 ```
   const vars1 = [
    {
      eventTypeId: 'eventTypeId',
      variablesJson: { x: 1, y: 2 },
    },
  ];

  const vars2 = [
    {
      eventTypeId: 'eventTypeId',
      variablesJson: { x: 1, z: 3 },
    },
  ];

  client.publishFusionMessage(jwt, vars1);
  client.publishFusionMessage(jwt, vars2);
  ```

hans-notifi:
> @hans-notifi The reason you are not able to input `{x: 1, z: 2}` is that you explicitly define the generic type in `FusionMessage`. If we do not define the generic type, it will fall back to `object` type and allow whatever shape we want as long as it is an `object`. example:
> 
> ```
>   const vars1 = [
>    {
>      eventTypeId: 'eventTypeId',
>      variablesJson: { x: 1, y: 2 },
>    },
>  ];
> 
>  const vars2 = [
>    {
>      eventTypeId: 'eventTypeId',
>      variablesJson: { x: 1, z: 3 },
>    },
>  ];
> 
>  client.publishFusionMessage(jwt, vars1);
>  client.publishFusionMessage(jwt, vars2);
> ```

The example I posted isn't from the actual code - it's what I would prefer the code to ideally do if we're enforcing type checking. My point is that unless `publishFusionMessage` _itself_ is made to take generic type parameters of what the message should be shaped like, applying any sort of generic type bounds to the `FusionMessage` type is useless.

eric-notifi (approved):
LGTM! thanks Hans

PR #478, opened 2024-02-20
Add new EVM chains

ghost:
Add support for new EVM chains, need to wait for services PR to be merged to re-generate proto stubs

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -101,7 +101,6 @@ export type WalletWithSignMessage =
     | 'OPTIMISM'
     | 'ZKSYNC'
     | 'BASE'
-    | 'INJECTIVE'
     | 'BLAST'
     | 'CELO'
     | 'MANTLE'
@@ -143,7 +142,13 @@ export type WalletWithSignMessage =
     accountAddress: string;
     walletPublicKey: string;
     signMessage: Uint8SignMessageFunction;
-  }>;
+  }>
+  | Readonly<{
+    walletBlockchain: 'INJECTIVE';
+    accountAddress: string;
+    walletPublicKey: string;
+    signMessage: Uint8SignMessageFunction;
+  }>
 
 export type ConnectWalletParams = Readonly<{
   walletParams: WalletWithSignParams;


packages/notifi-frontend-client/lib/client/ensureSource.ts
```diff
@@ -210,6 +210,22 @@ const ensureWalletBalanceSources = async (
           return 'ETHEREUM_WALLET';
         case 'BASE':
           return 'ETHEREUM_WALLET';
+        case 'BLAST':
+          return 'BLAST_WALLET';
+        case 'CELO':
+          return 'CELO_WALLET';
+        case 'INJECTIVE':
+          return 'INJECTIVE_WALLET';
```
eric-notifi, line 218: Seems `INJECTIVE_WALLET` is still not yet in `enum SourceType`. This will causes the typescript compile error

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -73,48 +81,56 @@ export type WalletWithSignParams = Readonly<{
...
+    walletBlockchain:
+    | 'ETHEREUM'
+    | 'POLYGON'
+    | 'ARBITRUM'
+    | 'AVALANCHE'
+    | 'BINANCE'
+    | 'OPTIMISM'
+    | 'ZKSYNC'
+    | 'BASE'
+    | 'INJECTIVE'
```
eric-notifi, line 98: Double confirm wether it is good to put injective here?
Seems that we also need to have `accountAddress` for INJECTIVE. don't we?

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -264,6 +280,14 @@ export class NotifiFrontendClient {
 
     let loginResult: Types.UserFragmentFragment | undefined = undefined;
     switch (walletBlockchain) {
+      case 'INJECTIVE':
```
eric-notifi, line 283: Not sure should INJECTIVE belong to here.
Seems INJECTIVE requires publickkey but also accountId.

eric-notifi (changes requested):
There are some nits listed below. 
And I think we might also want to add supported chains' param types into `packages/notifi-react-card/lib/context/NotifiContext.tsx` ex.

```typescript
export type XXXXParams = Readonly<{
  walletBlockchain: 'XXXX';
  walletPublicKey: string;
  signMessage: Uint8SignMessageFunction;
}>;
```

packages/notifi-frontend-client/lib/client/ensureSource.ts
eric-notifi: I believe it is still not in prod since I cannot find `INJECTIVE_WALLET` any using prod playground (https://api.notifi.network/gql/)

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: Shall we also revert the Injective here??

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -345,19 +375,9 @@ export class NotifiFrontendClient {
         );
         return signature;
       }
+      case 'INJECTIVE':
       case 'OSMOSIS':
       case 'ZKSYNC':
```
eric-notifi, line 380: I think this (removing entire section here) might break the login section since the following 3 chains  is the type of `NotifiConfigWithPublicKeyAndAddress` wich does not have `walletPublicKey` field. 
- `OSMOSIS`
- `ZKSYNC`
- `INJECTIVE`

eric-notifi (changes requested):
We might also want to add those new chains in this file:
`packages/notifi-react-card/lib/context/NotifiContext.tsx`

eric-notifi (changes requested):

eric-notifi (commented):
Line 825 in `packages/notifi-react-card/lib/hooks/useNotifiSubscribe.ts`   `await client.connectWallet(params);` has type conflict, might need to have a look

packages/notifi-react-hooks/lib/hooks/useNotifiClient.ts
```diff
@@ -126,6 +126,8 @@ const signMessage = async ({
   timestamp: number;
 }>): Promise<string> => {
   switch (params.walletBlockchain) {
+    case 'XION':
+      throw new Error('XION not supported with react-hooks, please migrate to Notifi Client.');
```
eric-notifi, line 130: the term `Notifi Client` might be confusing to users. 
I would use `XION not supported with react-hooks, please enable frontend-client (isUsingFrontendClient={true}`.

eric-notifi:
@jamie-notifi Looks like injective login is working with this PR. However, the Metamask sign is still not available.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -251,6 +293,74 @@ export class NotifiFrontendClient {
...
+    }
+    if (checkIsConfigWithDelegate(this._configuration)) {
+      const { delegatedAddress, delegatedPublicKey, delegatorAddress } = this._configuration;
+      const { nonce } = await this.beginLogInWithWeb3({
+        authAddress: delegatorAddress,
+        authType: 'COSMOS_AUTHZ_GRANT',
+      })
+
+      const message = `${SIGNING_MESSAGE}${nonce}}`;
+      const params = { walletBlockchain: 'XION', message, signMessage: signMessageParams.signMessage } as const
```
eric-notifi, line 315: Small nit? Why prevent we using `as SignMessageParams`

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -251,6 +293,74 @@ export class NotifiFrontendClient {
...
+        nonce,
+        signature,
+        signedMessage: message,
+        signingAddress: delegatedAddress,
+        signingPubkey: delegatedPublicKey
+      })
+      user = completeLogInWithWeb3.user
+
+    }
+    else if (checkIsConfigWithPublicKeyAndAddress(this._configuration)) {
```
eric-notifi, line 331: Do we suppose to have this `else if` case?
As far as I know from the code, it case will never be hit as long as the walletBlockchain is 'XION`. Not sure if I am wrong?

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -251,6 +293,74 @@ export class NotifiFrontendClient {
     };
   }
 
+  private async logInWithWeb3(
```
eric-notifi, line 296: I am just wondering the naming role...
It seems like only 'XION` is using this method, but we name it as with `web3` which is kind of general name. Doest it mean we will use it for other upcoming supported chain in the future?

eric-notifi (approved):
LGTM, prod tested. Both INJ login works (both metamask / Keplr)

PR #480, opened 2024-02-23
OpenDefi/MVP-3964: Alert Subscription

eric-notifi:
Add Alert Subscription view and its related components
No need
MVP-3964

### DIFF
--- a/packages/notifi-dapp-example/src/app/globals.css
+++ b/packages/notifi-dapp-example/src/app/globals.css
@@ -5,12 +5,13 @@
 @layer base {
   :root {
     /* Notifi global color */
-    --notifi-grey0: 245 246 241;
+    --notifi-gray0: 245 246 241;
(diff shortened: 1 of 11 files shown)


packages/notifi-dapp-example/src/app/globals.css
eric-notifi, line 1: The hook name here should be `useNotifiSubscribeAlert`. (Will rename it in the upcoming ticket: MVP-3970)

packages/notifi-dapp-example/src/hooks/useSubscribeAlert.ts
eric-notifi, line 1: The hook name here should be `useNotifiSubscribeAlert`. (Will rename it in the upcoming ticket: MVP-3970)

marukohao (approved):
LGTM

PR #483, opened 2024-02-28
OpenDefi/MVP-4108+4109: Wallet connect/ expiry page redesign

eric-notifi:
    - Wallet page redesign
    - Expired jwt page redesign
    - optimize global state
No need

MVP-4108/MVP-4109


marukohao (approved):

packages/notifi-dapp-example/src/context/GlobalStateContext.tsx
```diff
@@ -16,23 +16,51 @@ import {
...
 export const GlobalStateContextProvider: FC<PropsWithChildren> = ({
   children,
 }) => {
   const [isGlobalLoading, setIsGlobalLoading] = useState(false);
+
   const [globalError, setGlobalError] = useState<string | null>(null);
+  const [globalInfoModal, setGlobalInfoModal] =
+    useState<GlobalInfoModal | null>(null);
 
+  const popGlobalInfoModal = (modalData: Partial<GlobalInfoModal>) => {
```
marukohao, line 54: just be curious, why we use the popGlobalInfoModal here but not only use setGlobalInfoModal for showing the global info

eric-notifi: the name start with `set` will be normally considered as native react hook useState with type of `Dispatch`. To avoid that, we name it with `pop` to represent the action that we want to populate the modal CC: @marukohao

PR #482, opened 2024-02-27
Dashboard destination and edit destination section

marukohao:
- [ ] Describe the purpose of the change
add destination section on dashboard
create edit destination modal
add useNotifiTargets hook and refactor the save target logic

- [ ] Create test cases for your changes if needed

- [ ] Include the ticket id if possible (Collaborator only)
https://notifi.atlassian.net/jira/software/c/projects/MVP/boards/1?selectedIssue=MVP-3963

PR #484, opened 2024-02-28
Ankush/mvp 4086

ankush-notifi:

https://notifi.atlassian.net/browse/MVP-4086

### DIFF
--- a/packages/notifi-dapp-example/src/components/Toggle.tsx
+++ b/packages/notifi-dapp-example/src/components/Toggle.tsx
@@ -4,14 +4,12 @@ export type ToggleProps = Readonly<{
   disabled: boolean;
   checked: boolean;
   onChange?: () => void;
-  handleChange?: (val: boolean) => void;
 }>;
 
 export const Toggle: React.FC<ToggleProps> = ({
   disabled,
   checked,
   onChange,
-  handleChange,
 }: ToggleProps) => {
   return (
     <label
@@ -25,8 +23,7 @@ export const Toggle: React.FC<ToggleProps> = ({
         className="sr-only peer"
         checked={checked}
         disabled={disabled}
-        onChange={(e) => {
-          handleChange?.(e.target.checked);
+        onChange={() => {
           onChange?.();
         }}
       />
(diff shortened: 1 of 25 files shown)


packages/notifi-dapp-example/src/components/Toggle.tsx
```diff
@@ -4,12 +4,14 @@ export type ToggleProps = Readonly<{
...
   checked: boolean;
   onChange?: () => void;
+  handleChange?: (val: boolean) => void;
 }>;
 
 export const Toggle: React.FC<ToggleProps> = ({
   disabled,
   checked,
   onChange,
+  handleChange,
```
eric-notifi, line 14: Do we need an additional handle change? 
Looks like we can simply utilize `onChange`

eric-notifi (changes requested):
Generally looks nice. Some small nits herewith.
It seems going to have some conflicts with [this PR](https://github.com/notifi-network/notifi-sdk-ts/pull/482). There are some changes according to the repo convention. Suggest to take a look at that PR and Rebase that one.
And looks like the real test in my local env (dev tenant) is not working well. Do you run into the same issue (check the video herewith).

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/f5c92cf3-54f6-484d-bef8-3017ec4f39e0

And the github action seems not going through. might need to take a look. package-lock.json issue?

@ankush-notifi @marukohao

packages/notifi-dapp-example/src/components/Toggle.tsx
```diff
@@ -23,7 +23,9 @@ export const Toggle: React.FC<ToggleProps> = ({
         className="sr-only peer"
         checked={checked}
         disabled={disabled}
-        onChange={onChange}
+        onChange={() => {
+          onChange?.();
```
marukohao, line 27: Why do we need to use optional chaining here?

marukohao:
Slack works good for sign up, when edit destinations, toggle on then toggle off the slack, get an error message 'Failed to save", and when check on the networks, seems we didn't call the updateTarget query.
Screenshot 2024-02-29 at 1 46 51 PM

Untitled

packages/notifi-dapp-example/src/components/Toggle.tsx
ankush-notifi: since `onChange` is optional in props, so It's safe to use optional chaining here.

marukohao:
LGTM, can you fix the action fail before merging?

eric-notifi (approved):
LGTM

PR #488, opened 2024-03-05
Temp fix for Implementing Passover for Alert Filter (WIP)  

mayuranganesathas:

### DIFF
--- a/packages/notifi-core/lib/models/FusionEventDescriptor.ts
+++ b/packages/notifi-core/lib/models/FusionEventDescriptor.ts
@@ -1,0 +1,5 @@
+export type FusionEventDescriptor = {
+  id: string;
+  name: string;
+  metadata: string;
+};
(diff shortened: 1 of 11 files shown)


packages/notifi-core/lib/models/FusionEventDescriptor.ts
```diff
@@ -0,0 +1,5 @@
+export type FusionEventDescriptor = {
```
eric-notifi, line 1: I think we can just simply use the auto generated type `Types.FusionEventDescriptor`, cant we?

packages/notifi-core/lib/models/TenantConfig.ts
```diff
@@ -13,4 +15,5 @@ export type TenantConfig = Readonly<{
   id: string | null;
   type: string;
   dataJson: string | null;
+  fusionEvent: Array<FusionEventDescriptor>;
```
eric-notifi, line 18: Same as above

packages/notifi-react-card/lib/components/subscription/EventTypeFusionRenderer.tsx
eric-notifi, line 1: We might want to add an `TODO` section so that this section requires further refactor.

eric-notifi (changes requested):

PR #490, opened 2024-03-06
Fix the font family and update destination card shadow

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/src/app/page.tsx
+++ b/packages/notifi-dapp-example/src/app/page.tsx
@@ -6,11 +6,8 @@ import { PoweredByNotifi } from '@/components/PoweredByNotifi';
 import { useRouterAsync } from '@/hooks/useRouterAsync';
 import { useChain, useWalletClient } from '@cosmos-kit/react';
 import '@interchain-ui/react/styles';
-import { Inter } from 'next/font/google';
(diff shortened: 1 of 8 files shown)


packages/notifi-dapp-example/src/app/page.tsx
```diff
@@ -28,7 +25,7 @@ export default function Home() {
   }, [client]);
 
   return (
-    <main className="flex min-h-screen flex-col items-center justify-center inter.className">
+    <main className="flex min-h-screen flex-col items-center justify-center font-sans">
```
eric-notifi, line 28: I am not sure what prevent us using the original way like:
```typescript
const inter = Inter({ subsets: ['latin'] });
// ...
<main className={`flex min-h-screen flex-col items-center justify-center ${inter.className}`}>
```

marukohao: I have tried and this can't apply the font family inter

eric-notifi: Yeah, I think you cardcoded that string as "inter.className", but we actually have to use the value in that variable like below. This is working in my end.
Screenshot 2024-03-07 at 10 27 48

eric-notifi (approved):
LGTM

PR #495, opened 2024-03-12
fix ui feedbacks

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
+++ b/packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
@@ -8,9 +8,9 @@ export default function NotifiFTU({ children }: { children: React.ReactNode }) {
   useNotifiRouter();
 
   return (
-    <div className="w-full h-screen flex flex-col items-center justify-center bg-notifi-page-bg">
-      <div className="fixed top-8 left-8 right-8 flex justify-between">
-        <div className="left-8 flex items-center">
+    <div className="w-full flex flex-col items-center justify-center bg-notifi-page-bg">
+      <div className="w-full mt-8 flex flex-row justify-between">
+        <div className="ml-8 flex items-center">
           <Image
             src="/logos/injective.png"
             width={115}
@@ -23,7 +23,7 @@ export default function NotifiFTU({ children }: { children: React.ReactNode }) {
             INJECTIVE ECOSYSTEM ALERTS
           </div>
         </div>
-        <div className=" p-2 bg-white rounded-lg h-7">
+        <div className=" p-2 bg-white rounded-lg h-7 mr-8">
           <PoweredByNotifi />
         </div>
       </div>
--- a/packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
+++ b/packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
@@ -8,7 +8,7 @@ export default function NotifiFTU({ children }: { children: React.ReactNode }) {
   useNotifiRouter();
 
   return (
-    <div className="w-full flex flex-col items-center justify-center bg-notifi-page-bg">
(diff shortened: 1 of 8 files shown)


packages/notifi-dapp-example/src/components/DashboardSideBar.tsx
```diff
@@ -83,7 +83,7 @@ export const DashboardSideBar: FC<DashboardSideBarProps> = ({
                 cardView === 'alertSubscription'
                   ? 'text-notifi-icon-selected'
                   : 'text-notifi-icon-unselected'
-              }`}
+              } mt-[2px]`}
```
eric-notifi, line 86: We better use "rem" instead of 'px' so that it can auto adjust when users zoom in/out the browser.

packages/notifi-dapp-example/src/components/HistoryList.tsx
```diff
@@ -71,7 +71,7 @@ export const HistoryList: React.FC<HistoryListProps> = ({
           />
         </div>
       ) : null}
-      <div className="min-h-0 overflow-scroll grow">
+      <div className="min-h-0 grow">
```
eric-notifi, line 74: This will also make "inbox" row scrollable. Not sure if it is a expected behavior?

packages/notifi-dapp-example/src/components/AlertSubscription.tsx
```diff
@@ -18,15 +18,11 @@ export const AlertSubscription: React.FC<AlertSubscriptionRowProps> = ({
...
-      className={`flex flex-col items-center 2xl:px-[15.75rem] xl:px-[10rem] ${
-        title ? 'h-[94vh]' : ''
-      }`}
-    >
+    <div className="flex flex-col items-center 2xl:px-[15.75rem] xl:px-[10rem] min-h-0 grow h-full">
       {title ? (
         <div className="mt-8 mb-6 font-medium text-lg">{title}</div>
       ) : null}
-      <div className="flex flex-wrap gap-4 justify-center overflow-scroll">
+      <div className="flex flex-wrap gap-4 justify-center w-72 md:w-[37rem]">
```
eric-notifi, line 25: OH ... So we even not allow to  have y axis scroll

packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
eric-notifi, line 1: Realized the loading style while moving from 'connect wallet' page to 'notifi related page' has broken some how. FYI below.

Screenshot 2024-03-13 at 9 23 48


https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/4cbb0b38-152f-4075-8753-8289892ab75b

eric-notifi, line 1: This might be out of scope of feedback fix, but I realized that We might not need to wrap  the `{children}` in `packages/notifi-dapp-example/src/app/notifi/layout.tsx` since it is already inside `<NotifiContext/>`. 
Meaning as long as those components are wrappedwith `<NotifiContext />` should already be wrapped with <NotifiFormProvider/>`. We might not want to use this context provider individually.
@marukohao

packages/notifi-dapp-example/src/components/AlertSubscription.tsx
marukohao: Hmm, I think so, we have the vertical line break too

packages/notifi-dapp-example/src/components/HistoryList.tsx
marukohao: Ah, you are right,  have updated this to overflow-y-scroll as Alex and Kai has the horizontal line in their alert history view, which I can't reproduce

packages/notifi-dapp-example/src/app/notifi/ftu/layout.tsx
marukohao: fixed

PR #509, opened 2024-03-26
OpenDefi/mvp4251-2: Optimization

eric-notifi:
We might want to avoid using the context of card as possible.
IMO, we can just use `isGlobalLoading` state instead of the `loading` state imported from card.

MVP-4251

### DIFF
--- a/packages/notifi-dapp-example/src/components/ConfigDestinationModal.tsx
+++ b/packages/notifi-dapp-example/src/components/ConfigDestinationModal.tsx
@@ -1,4 +1,5 @@
 import { LoadingSpinner } from '@/components/LoadingSpinner';
+import { useGlobalStateContext } from '@/context/GlobalStateContext';
 import { useNotifiTargetListener } from '@/hooks/useNotifiTargetListener';
 import {
   CardConfigItemV1,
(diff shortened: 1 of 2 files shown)


marukohao (approved):

packages/notifi-dapp-example/src/components/ConfigDestinationModal.tsx
```diff
@@ -29,7 +30,7 @@ export const ConfigDestinationModal: React.FC<ConfigDestinationModalProps> = ({
           </p>
         </div>
         <div className="flex flex-col items-center justify-center">
-          {loading ? (
+          {isGlobalLoading ? (
```
marukohao, line 33: Should we remove the LoadingSpinner here? as we already have the loadingSpinner when isGlobalLoading is true

marukohao:
Seems the loading spinner after when subscribing is missing
Untitled

eric-notifi:
Ah ...., I think we are abusing the `globalLoading`. 
When doing sign-up, multi `setIsGlobalLoading`s are called, which causes issue. It gets set to `false` first in FTU page. 
I think we might not want to use the global Loading here. 
Instead, we want to use the loading in `useNotifiTopics` as the loading is from subscribing alerts. 

> We can not fix it now because I am working on improving loading speed by modifying these files. So I will fix it in another PR. 

CC: @marukohao

PR #513, opened 2024-03-29
add leap and phantom wallet

marukohao:


packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  const phantomSignArbitrary = useCallback(
+    async (message: string) => {
+      if (!walletKeysPhantom || !injectivePhantomWallet.strategies.phantom) {
+        return;
+      }
+      setIsLoading(true);
+      try {
+        const result = await injectivePhantomWallet.signArbitrary(
+          walletKeysPhantom.bech32,
```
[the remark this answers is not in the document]
eric-notifi: Phanton should be following the same interface as metamask. so the wallet key might be hex & bech32. Not base64 & bech32

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+          message,
```
eric-notifi, line 154: Just FYI, in metamask, we convert the message to hex by `Buffer.from(message).toString('hex')`. 
But I tried doing so, the `signiture not match` error is still happening.

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
eric-notifi, line 1: I think we might also want to implement the following two features:
1. local storage so that users will be re-connected automaticly 
2. the account change feature, so it will refresh data when account changed

packages/notifi-dapp-example/src/utils/localStorageUtils.ts
eric-notifi, line 1: We might want to reuse the localStorageUtils.ts in wallet-provider package

packages/notifi-wallet-provider/lib/context/NotifiWallets.tsx
```diff
@@ -73,7 +73,11 @@ export const NotifiWalletProvider: React.FC<PropsWithChildren> = ({
 
   useEffect(() => {
     const storageWallet = getWalletsFromLocalStorage();
-    if (storageWallet) {
+    if (
+      storageWallet &&
+      selectedWallet &&
+      Object.keys(wallets).includes(selectedWallet)
```
eric-notifi, line 79: Do we have to have these two extra condition?

packages/notifi-dapp-example/src/context/NotifiContext.tsx
```diff
@@ -15,40 +17,86 @@ export const NotifiContextProvider: React.FC<PropsWithChildren> = ({
...
+  const { wallets: injectiveWallets, selectedWallet: injectiveSelectedWallet } =
+    useInjectiveWallets();
 
   if (
-    !selectedWallet ||
-    !wallets[selectedWallet].walletKeys ||
-    !wallets[selectedWallet].signArbitrary
+    (!selectedWallet ||
+      !wallets[selectedWallet].walletKeys ||
+      !wallets[selectedWallet].signArbitrary) &&
```
eric-notifi, line 26: Do we need to use the `&&` statement? or we can simply chain all those conditions with `||` ?

packages/notifi-dapp-example/src/context/NotifiContext.tsx
eric-notifi, line 1: Do you get the type error here `signMessage={signMessage}`? 
Or it only happens in my env?

packages/notifi-dapp-example/src/utils/types.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @param hex eth hex address starting with 0x
+ */
+export type WalletKeysBase = {
+  bech32: string;
+  base64: string;
+  hex: string;
+};
+
+export type LeapWalletKeys = PickKeys<WalletKeysBase, 'bech32' | 'base64'>;
+export type PhantomWalletKeys = PickKeys<WalletKeysBase, 'bech32' | 'base64'>;
```
eric-notifi, line 17: Here is the same mentioned above

eric-notifi (changes requested):

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
marukohao: Should we convert message to hex for phantom?

packages/notifi-dapp-example/src/context/NotifiContext.tsx
marukohao: yes, you are right, we can chain all the conditions with 'or'

packages/notifi-wallet-provider/lib/context/NotifiWallets.tsx
marukohao: yes, if not adding      
` selectedWallet && Object.keys(wallets).includes(selectedWallet)`
`wallets[storageWallet.walletName]` will get error as, the wallet name is not in wallets if we connect injective wallet

packages/notifi-dapp-example/src/utils/localStorageUtils.ts
marukohao: I think we may need to separate localStarageUtils, as we are using the type of wallet here and we separate the type of wallet for our existing wallet and injective wallet.

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+    phantom: new PhantomWallet(
+      injectivePhantomWallet.strategies.phantom ? true : false,
+      walletKeysPhantom,
+      phantomSignArbitrary,
+      connectPhantomWallet,
+      disconnectPhantomWallet,
+    ),
+  };
+
+  useEffect(() => {
```
marukohao, line 220: for this feature(local storage so that users will be re-connected automaticly) you mentioned that we need to add, do you mean this section below?

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
eric-notifi: No, I think the `getAddresses` from the `walletStrategy` is already an array of hex string

packages/notifi-dapp-example/src/utils/localStorageUtils.ts
eric-notifi: That is very true. It looks kind of confusing tho. shall we :
1. Make a TODO comment saying we want to move it to wallet-provider package
2. Rename it to setInjWalletKeysToLocalStroage`

packages/notifi-wallet-provider/lib/context/NotifiWallets.tsx
eric-notifi: I see, this looks like a kind of a patch workaround tho. 
Shall we add a comment saying the purpose of these two lines and also todo to remove that

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
```diff
@@ -0,0 +1,249 @@
...
+    const storageWallet = getWalletsFromLocalStorage();
+    if (
+      storageWallet &&
+      selectedWallet &&
+      Object.keys(wallets).includes(selectedWallet)
+    ) {
+      const walletName = storageWallet.walletName;
+      (wallets as { [key: string]: any })[storageWallet.walletName]
```
eric-notifi, line 228: Do we need to cast the `[key: string]: any` type?

packages/notifi-dapp-example/src/context/InjectiveWalletContext.tsx
eric-notifi: Ah .... right. But it seems not re-login.

https://github.com/notifi-network/notifi-sdk-ts/assets/127958634/2d0ad387-a2f0-4c33-893a-6ce98aa1e55f

packages/notifi-dapp-example/src/context/NotifiContext.tsx
marukohao: Yeah, the type error is exist on my end, don't know how to fix it lol

eric-notifi (approved):
LGTM (Also tested can sign the msg 👍)

> Just a note, that we might want to someday in the future move the wallets to our own package to
> - boot the performance
> - increase the scalability
> - Remove the redundant codes.

PR #517, opened 2024-04-05
fix wallet.isInstalled does not work

marukohao:
injectiveWallet.strategies.leap and injectiveWallet.strategies.phantom still returns true event if the browser haven't installed the extension
add leap and phantom wallets to MobilePromptModalProps


eric-notifi (approved):
LGTM

PR #520, opened 2024-04-08
Add slack webhook

kai-notifi:


marukohao:
The cypress test failed, I think we may need to update the generated.ts file

kai-notifi:
> The cypress test failed, I think we may need to update the generated.ts file

Issue was cypress was using prod for its gql types. New changes has been pushed out to fix

PR #525, opened 2024-04-10
Added implementation of StateChangedSubscription 

vaishali-notifi:

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -940,6 +940,17 @@ export class NotifiFrontendClient {
     return result;
   }
 
+  async stateChanged(): Promise<
+    Types.StateChangedSubscription['stateChanged']
+  > {
+    const query = await this._service.stateChanged({})
+    const result = query.stateChanged;
+    if (!result) {
+      throw new Error('Failed to state changes');
+    }
+    return result;
+  }
+
   async getUserSettings(): Promise<Types.GetUserSettingsQuery['userSettings']> {
     const query = await this._service.getUserSettings({});
     const result = query.userSettings;
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -30,6 +30,9 @@ import {
   ensureTelegram,
   ensureWebhook,
 } from './ensureTarget';
+import { Client } from 'graphql-ws';
+import { NotifiSubscriptionClient } from './NotifiSubscriptionClient';
+import { StateChangedNotifiService } from './StateChangedNotifiService';
 
 export type SignMessageParams =
   | Readonly<{
@@ -219,6 +222,7 @@ export class NotifiFrontendClient {
     private _configuration: NotifiFrontendConfiguration,
     private _service: NotifiService,
     private _storage: NotifiStorage,
+    private _wsUrl: string
   ) { }
 
   private _clientRandomUuid: string | null = null;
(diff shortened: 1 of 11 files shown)


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -219,6 +222,7 @@ export class NotifiFrontendClient {
     private _configuration: NotifiFrontendConfiguration,
     private _service: NotifiService,
     private _storage: NotifiStorage,
+    private _wsUrl: string
```
eric-notifi, line 225: I would not put the wsUrl here since it is the constant. 
Now all the endpoint related operations are in `notifi-graphql`'s `GraphQLClient`. Do you think would it be a good idea to move it here?

packages/notifi-frontend-client/lib/client/NotifiSubscriptionClient.ts
eric-notifi, line 1: Well, we might not want to create multiple `cleint` object here. As mentioned, most of our graphql endpoint operations are located in `notifi-graphql`'s `NotifiService`. Do you think it is good idea to move it over there?

packages/notifi-frontend-client/lib/client/StateChangedNotifiService.ts
eric-notifi, line 1: The same comment as above

packages/notifi-frontend-client/lib/client/clientFactory.ts
```diff
@@ -32,5 +34,6 @@ export const newFrontendClient = (args: ConfigFactoryInput) => {
   const config = newFrontendConfig(args);
   const service = newNotifiService(config);
   const storage = newNotifiStorage(config);
-  return new NotifiFrontendClient(config, service, storage);
+  const wsurl = envWSGraphqlUrl(config.env);
+  return new NotifiFrontendClient(config, service, storage, wsurl);
```
eric-notifi, line 38: Similar comment as above. 
If we migrate all the gql related operations to `notifi-graphql` package, we will no longer need to have input the wsrul here I believe. How do you think

packages/notifi-frontend-client/lib/configuration/NotifiFrontendConfiguration.ts
```diff
@@ -196,10 +196,23 @@ export const envUrl = (env: NotifiEnvironment): string => {
...
+    case 'Development':
+      return 'ws://api.dev.notifi.network/gql';
+    case 'Local':
+      return 'ws://localhost:5001/gql';
+    case 'Production':
+      return 'ws://api.notifi.network/gql';
+    case 'Staging':
+      return 'ws://api.stg.notifi.network/gql';
+  }
+};
```
eric-notifi, line 218: I think it will be more efficient to combine this section with `envUrl`. Something like below:

```typescript
export const envUrl = (env: NotifiEnvironment, endpointType?: 'websocket' | 'http'): string => {
  switch (env) {
    case 'Development':
      return endpointType === 'websocket' ? 'ws://api.dev.notifi.network/gql' :  'https://api.dev.notifi.network/gql' : '';
    //  others ....
  }
};
```

eric-notifi (changes requested):

packages/notifi-graphql/lib/gql/queries/stateChanged.gql.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from 'graphql-request';
+
+export const stateChanged = gql`
+subscription stateChanged {
+  stateChanged {
```
eric-notifi, line 5: Not sure if this endpoint available? I cannot find it either in prod or env. 
So it might cause build fail.

PR #533, opened 2024-04-15
notifi-react/MVP-4342: finalize context & rename & example case

eric-notifi:
  - Impl use case of NotifiContext in example package
  - context usage example case & Refactor

No need
MVP-4342


marukohao (approved):
LGTM

PR #535, opened 2024-04-17
Add StateChangedNotifiService

vaishali-notifi:

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -988,9 +988,9 @@ export class NotifiFrontendClient {
     return result;
   }
 
-  async subscribeNotificationHistoryStateChanged(userid: string, onHistoryChanged: () => void): Promise<void> {
+  async subscribeNotificationHistoryStateChanged(onHistoryChanged: () => void): Promise<void> {
     const token = await this._storage.getAuthorization();
-    StateChangedNotifiService(this._wsUrl, token?.token, userid, onHistoryChanged);
+    StateChangedNotifiService(this._wsUrl, token?.token, onHistoryChanged);
   }
 
   async getUserSettings(): Promise<Types.GetUserSettingsQuery['userSettings']> {
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -32,8 +32,6 @@ import {
   ensureTelegram,
   ensureWebhook,
 } from './ensureTarget';
-import { Client } from 'graphql-ws';
-import { NotifiSubscriptionClient } from './NotifiSubscriptionClient';
 import { StateChangedNotifiService } from './StateChangedNotifiService';
 
 export type SignMessageParams =
@@ -990,9 +988,9 @@ export class NotifiFrontendClient {
     return result;
   }
 
-  async subscribeNotificationHistoryStateChanged(onHistoryChanged: () => void): Promise<void> {
+  async subscribeNotificationHistoryStateChanged(userid: string, onHistoryChanged: () => void): Promise<void> {
     const token = await this._storage.getAuthorization();
-    StateChangedNotifiService(this._wsUrl, token?.token, onHistoryChanged);
+    StateChangedNotifiService(this._wsUrl, token?.token, userid, onHistoryChanged);
   }
 
   async getUserSettings(): Promise<Types.GetUserSettingsQuery['userSettings']> {
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -2,18 +2,19 @@ import { Types } from '@notifi-network/notifi-graphql';
 import { NotifiService } from '@notifi-network/notifi-graphql';
 
 import {
-  checkIsConfigWithDelegate,
-  checkIsConfigWithPublicKeyAndAddress,
   type NotifiConfigWithPublicKey,
   type NotifiConfigWithPublicKeyAndAddress,
   type NotifiFrontendConfiguration,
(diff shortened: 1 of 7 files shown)


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -204,25 +207,26 @@ export type SupportedCardConfigType = CardConfigItemV1;
...
+  }
 >;
 
 export class NotifiFrontendClient {
   constructor(
     private _configuration: NotifiFrontendConfiguration,
     private _service: NotifiService,
     private _storage: NotifiStorage,
-  ) {}
+    private _wsUrl: string
```
nimesh-notifi, line 228: nit: _wssUrl

packages/notifi-frontend-client/lib/client/NotifiSubscriptionClient.ts
```diff
@@ -0,0 +1,30 @@
...
+  });
+
+  client.on('connected', () => {
+    console.log('WebSocket connection opened');
+  });
+
+  client.on('error', (error) => {
+    console.error('WebSocket error:', error);
+  });
+
```
nimesh-notifi, line 19: you need handler for:
 client.on('close', () => {
  });
  
  in this, you should check to see if we're still authenticated, and if so, reconnect after 5 seconds

packages/notifi-frontend-client/lib/client/clientFactory.ts
```diff
@@ -5,6 +5,7 @@ import {
   ConfigFactoryInput,
   NotifiFrontendConfiguration,
   envUrl,
+  envWSGraphqlUrl,
```
nimesh-notifi, line 8: nit: we should use WSS instead of WS naming pattern

packages/notifi-frontend-client/lib/configuration/NotifiFrontendConfiguration.ts
```diff
@@ -203,3 +203,16 @@ export const envUrl = (env: NotifiEnvironment): string => {
       return 'https://api.stg.notifi.network/gql';
   }
 };
+
+export const envWSGraphqlUrl = (env: NotifiEnvironment): string => {
+  switch (env) {
+    case 'Development':
+      return 'ws://api.dev.notifi.network/gql';
+    case 'Local':
+      return 'ws://localhost:5001/gql';
```
nimesh-notifi, line 212: all of these should be wss instead of ws, except Local

nimesh-notifi: ws is equivalent to http
wss is essentially http**s**

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: I would not put the wsUrl here since it is the constant. 
Now all the endpoint related operations are in `notifi-graphql`'s `NotifiService`. Do you think would it be a good idea to move it here? So it will be similar like below:
``` typescript
import { createClient as createWsClient } from 'graphql-ws';
// ...
export const new NotifiService = (config: NotifiFrontendConfiguration) => {
 //  ...
  const wsUrl = envUrl(/* TODO, See my anohter comment */);
  const wsClient = createWsClient({ url });
  return new NotifiService(client, wsClient); // The endpoint operations to be migrated to 'notifi-graphql'
};
```

packages/notifi-frontend-client/lib/client/NotifiSubscriptionClient.ts
eric-notifi, line 1: Well, Similar comment as others. we might not want to create multiple `cleint` objects here. As mentioned, most of our graphql endpoint operations are located in `notifi-graphql`'s `NotifiService`. Do you think it is good idea to move it over there?

packages/notifi-frontend-client/lib/client/StateChangedNotifiService.ts
eric-notifi, line 1: The same comment as above

packages/notifi-frontend-client/lib/client/clientFactory.ts
```diff
@@ -32,5 +34,6 @@ export const newFrontendClient = (args: ConfigFactoryInput) => {
   const config = newFrontendConfig(args);
   const service = newNotifiService(config);
   const storage = newNotifiStorage(config);
-  return new NotifiFrontendClient(config, service, storage);
-};
+  const wsurl = envWSGraphqlUrl(config.env);
+  return new NotifiFrontendClient(config, service, storage, wsurl);
+};
```
eric-notifi, line 39: Similar comment as above.
If we migrate all the gql related operations to notifi-graphql package, we will no longer need to have input the wsrul here I believe. How do you think

packages/notifi-frontend-client/lib/configuration/NotifiFrontendConfiguration.ts
eric-notifi: I think it will be more efficient to combine this section with `envUrl`. Something like below:

```typescript
export const envUrl = (env: NotifiEnvironment, endpointType?: 'websocket' | 'http'): string => {
  switch (env) {
    case 'Development':
      return endpointType === 'websocket' ? 'wss://api.dev.notifi.network/gql' :  'https://api.dev.notifi.network/gql' : '';
    //  others ....
  }
};
```

eric-notifi (commented):
My idea is migrating the `endpoint` operation to `notifi-grapqhql`'s `Service` object. This way we have a `central service store` for all our gql operations.
`frontedn-client`'s an extra layer to wrap the `Service` introduced by `notifi-graphql` to create a new 'client' interface for users to call. So in this case IMO, frontendClient object should only call the service method provided by `notifi-graphql` like `_service.stateChanged` 

And the build seems failed somehow, might need to take a look. Not sure if it is caused by endpoint not yet available in prod ?

nimesh-notifi:
@vaishali-notifi can we close this pull request?

PR #537, opened 2024-04-17
Make NotifiEnvironment default to Production

hans-notifi:
'Production' is now the only setting for NotifiEnvironment that is used by external customers. Therefore, in all places where a NotifiEnvironment is passed (React card, frontend configuration, Node.js clients, etc.), make omitting the NotifiEnvironment parameter default to 'Production'.

### DIFF
--- a/packages/notifi-react-example/package.json
+++ b/packages/notifi-react-example/package.json
@@ -47,10 +47,10 @@
   },
   "scripts": {
     "start": "BROWSER=none react-app-rewired start",
-    "test": "cypress run --component --spec 'cypress/component/NotifiSubscriptionCard.cy.tsx' --browser firefox",
+    "test": "cypress run --component --spec 'cypress/component/NotifiSubscriptionCard.cy.tsx' --browser chrome",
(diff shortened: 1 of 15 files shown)


packages/notifi-react-example/package.json
```diff
@@ -47,10 +47,10 @@
   },
   "scripts": {
     "start": "BROWSER=none react-app-rewired start",
-    "test": "cypress run --component --spec 'cypress/component/NotifiSubscriptionCard.cy.tsx' --browser chrome",
+    "test": "cypress run --component --spec 'cypress/component/NotifiSubscriptionCard.cy.tsx' --browser firefox",
```
nimesh-notifi, line 50: was this change expected?

nimesh-notifi (approved):

packages/notifi-react-example/package.json
hans-notifi: Yeah, I had changed that because Chrome is not installed on my system. I just put in a commit to change it back.

PR #540, opened 2024-04-18
Integrate state changed graph ql subscription

vaishali-notifi:


packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -0,0 +1,82 @@
...
+      extensions: {
+        type: 'start'
+      }
+    },
+      {
+        next: (data) => {
+          if (actionOnMessageReceived) {
+            actionOnMessageReceived();
+          }
+          console.log("Subscription data recieved" + JSON.stringify(data));
```
eric-notifi, line 27: Instead of printing out the `data`, shall we put the data into the callback so that clients have more control? ex below
```typescript
onNext?: (data) => void
```

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -0,0 +1,82 @@
...
+        },
+        error: (error) => {
+          console.error('Subscription error:', error);
```
eric-notifi, line 30: The same as above, example, 
```typescript
onError?: (error) => void
```

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -0,0 +1,82 @@
...
+        },
+        complete: () => {
+          console.log('Subscription completed');
```
eric-notifi, line 33: Same
```ts
onComplete?: ()=>void
```

packages/notifi-react-card/lib/hooks/useUnreadState.ts
```diff
@@ -30,27 +30,20 @@ export const useUnreadState = () => {
...
-        /* Intentionally empty (Concurrent can only possibly happens here instead of inside interval) */
-      });
-    const interval = setInterval(() => {
-      if (!walletPublicKey || !isClientAuthenticated) return;
+    console.log("Called: useUnreadState.subscribeNotificationHistoryStateChanged for user:");
+    setIsStateChangedSubscribed(true);
+
+    frontendClient.subscribeNotificationHistoryStateChanged(() => {
+
+      console.log("CallBack: subscribeNotificationHistoryStateChanged");
```
eric-notifi, line 40: Not sure if this log is necessary for production?

packages/notifi-react-card/lib/hooks/useUnreadState.ts
```diff
@@ -30,27 +30,20 @@ export const useUnreadState = () => {
...
 
       frontendClient.getUnreadNotificationHistoryCount().then((res) => {
         const unreadNotificationCount = res.count;
         setUnreadNotificationCount(unreadNotificationCount);
       });
-    }, Math.floor(Math.random() * 5000) + 5000); // a random interval between 5 and 10 seconds to avoid spamming the server
-
-    return () => clearInterval(interval);
+    });
```
eric-notifi, line 46: Not sure if we want to close the websocket after the component is unmounted? Otherwise, will it cause multiple wss connections when the component is rendered multiple times?

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -0,0 +1,82 @@
...
+      }, 3000);
+    });
+
+    this.wsClient.on('message', (data) => {
+      if (data.type == 'next') {
+        console.log('Notification history has been changed: ', data);
+      } else {
+        console.log('Received data:', data);
+      }
+    });
```
eric-notifi, line 72: Similarly, shall we also exposes these:
```ts
onConnected?: () => void
onConnectionError?: (error)=> void
onConnectionClosed?: (event)=> void
onMessage: (data) => void // Not sure if it has any difference from the `.subscribe`' `next` argument
```

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
vaishali-notifi: done

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
vaishali-notifi: We are already handling at subscribe query level so do we really need to expose this also, it is kind of redundant and also initializing of client we are doing inside this class and also Wsclient is private.

eric-notifi (approved):
Generally LGTM. Just please make sure that the build will not fail before merging 
BTW, just a reminder. please remove the conflict PR as well

PR #548, opened 2024-05-01
MVP-4297-1-patch: context optimize

eric-notifi:
  - Add `subscriptionValueOrRef: ValueOrRef<string>;` in  `fusionMetadata`
  - Add `topicGroupName?: string;` in `fusionEventTypeItem`
  - Optimize TargetContext and TopicContext
No need
MVP-4297 milestone#1 context

### DIFF
--- a/packages/notifi-frontend-client/lib/models/FusionEvent.ts
+++ b/packages/notifi-frontend-client/lib/models/FusionEvent.ts
@@ -1,11 +1,14 @@
 import { Types } from '@notifi-network/notifi-graphql';
 
+import { ValueOrRef } from './SubscriptionCardConfig';
+
 export type FusionEventMetadata = {
   uiConfigOverride?: {
     topicDisplayName?: string;
     historyDisplayName?: string;
     icon?: Types.GenericEventIconHint;
     customIconUrl?: string;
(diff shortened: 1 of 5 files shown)


packages/notifi-frontend-client/lib/models/FusionEvent.ts
```diff
@@ -1,11 +1,14 @@
...
 
+import { ValueOrRef } from './SubscriptionCardConfig';
+
 export type FusionEventMetadata = {
   uiConfigOverride?: {
     topicDisplayName?: string;
     historyDisplayName?: string;
     icon?: Types.GenericEventIconHint;
     customIconUrl?: string;
+    subscriptionValueOrRef: ValueOrRef<string>;
```
marukohao, line 11: per our discussion, I think we also need to add 
`isSubscriptionValueSelectable?: boolean;`
here

eric-notifi: Thanks!!! you are right. I was misoriended. Let me change it to `isSubscriptionValueInputable?: boolean;` as it also could be be an input text field

marukohao (approved):

PR #549, opened 2024-05-01
Ankush/mvp 4347

ankush-notifi:
https://notifi.atlassian.net/browse/MVP-4347


packages/notifi-dapp-example/src/components/WalletSelectModal.tsx
```diff
@@ -32,34 +32,35 @@ export const WalletSelectModal: FC<WalletSelectModalProps> = ({
...
-          className="absolute right-8 top-8 text-gray-500 cursor-pointer"
-          onClick={() => setIsOpenWalletsModal(false)}
-        />
-        <div className="flex grow gap-6 p-5 justify-center items-center flex-wrap">
+
+        <div className="flex gap-6 px-12 pt-10 justify-center items-center flex-wrap">
           {objectKeys(wallets)
             .filter(
-              (wallet) => wallets[wallet].isInstalled && wallet === 'metamask',
+              (wallet) => wallets[wallet].isInstalled && wallet !== 'keplr',
```
marukohao, line 57: Is there a better way to sort out the wallets that we need for GMX instead of use wallet !== 'keplr'? as we may add more wallets in wallet provide in the future.

packages/notifi-wallet-provider/lib/hooks/useBinance.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (typeof window === 'undefined' || !window.BinanceChain) {
+    throw new Error(
+      `Cannot get ${walletName} without a window | Cannot get ${walletName} from window`,
+    );
+  }
+
+  if (window.BinanceChain) {
+    return window.BinanceChain;
+  } else if (document.readyState === 'complete') {
+    throw new Error(`Please install the ${walletName} extension`);
```
marukohao, line 173: We'll want to remove this error message per ODF.

ankush-notifi: yeah, I've taken care of that. The error message won't be visible on the frontend unless we specifically need it. 
this way, it remains available for troubleshooting.

eric-notifi (approved):

marukohao:
@ankush-notifi I'll merge this PR for now, please address feedbacks in following prs, thanks

packages/notifi-wallet-provider/lib/hooks/store.ts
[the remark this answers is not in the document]
ankush-notifi: Yes, I think we should remove it because it won't work as expected. Its logic relies on the window.ethereum object API, which can be overridden by another wallet's object if installed.

PR #554, opened 2024-05-06
Ankush/mvp 4508

ankush-notifi:
https://notifi.atlassian.net/browse/MVP-4508


packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -246,25 +276,44 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+        .catch((e) => setError(e as Error))
+        .finally(() => setIsLoading(false));
+    },
+    [frontendClient, targetGroupToBeSaved],
+  );
+
+  const renewToggleTargetGroup = useCallback(
+    async (
+      target: Extract<Target, 'discord' | 'slack' | 'wallet'>,
+      value: boolean,
```
eric-notifi, line 305: Rather than creating a new `renewToggleTargetGroup`, do you think it is possible to combine this one to existing `renewTargetGroup`? I am afraid exposing two interfaces might cause confusion.

As far as I concern, all the updated un-updated data is already in `targetGroupToBeSaved`, I would rather just add one more optional argument in `renewTargetGroup` to only renew single target if that is the case.

packages/notifi-graphql/lib/gql/fragments/Web3TargetFragment.gql.ts
```diff
@@ -0,0 +1,34 @@
...
+    name
+    accountId
+    walletBlockchain
+    targetProtocol
+    isConfirmed
+  }
+`;
+
+export const GetWeb3TargetResponseFragment = gql`
+  fragment GetWeb3TargetResponseFragment on Web3TargetsConnection {
```
eric-notifi, line 15: Just one question about the naming. @kai-notifi .
As far as know, this is the CB wallet target. Can I understand the intention of  naming it as Web3Target such a general name is that we will finally include all the blockchain related target into this one if there are come coming up??
CC: @ankush-notifi @nimesh-notifi

marukohao (approved):
LGTM

PR #557, opened 2024-05-08
AlertSubscription/AlertSubscriptionRow/AlertSubscriptionOptions updates

marukohao:
Update AlertSubscription/AlertSubscriptionRow/AlertSubscriptionOptions to support group topic

### DIFF
--- a/packages/notifi-dapp-example/src/components/AlertSubscription.tsx
+++ b/packages/notifi-dapp-example/src/components/AlertSubscription.tsx
@@ -11,24 +11,20 @@ export type AlertSubscriptionProps = {
   inFTU?: boolean;
 };
 
-export type AlertSubscriptionTopicRowCategory = 'standalone' | 'group';
+export type TopicRowCategory = 'standalone' | 'group';
 
-export type AlertSubscriptionTopicRowMetadata<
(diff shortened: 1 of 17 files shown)


packages/notifi-dapp-example/src/components/AlertSubscription.tsx
```diff
@@ -1,21 +1,77 @@
...
+> = T extends 'standalone'
+  ? AlertSubscriptionTopicStandaloneRowMetadata
+  : AlertSubscriptionTopicGroupRowMetadata;
+
+type TopicRowMetadataBase = {
+  index: number;
+};
+
+export type AlertSubscriptionTopicStandaloneRowMetadata =
+  TopicRowMetadataBase & {
```
eric-notifi, line 27: Just a nit, we might want to make names as consistent as possible since this is an "example" repo. So we do not want to confuse those devs using the repo. 
There are few naming here:
1. AlertSubscriptionTopicRow --> component name
2. AlertSubscriptionRow --> file name
3. TopicRow --> Sometimes show across the code base.
How do you think to make those name the same, could be simplified only `TopicRow`??

PR #561, opened 2024-05-09
Clear notification count on click of bell icon

vaishali-notifi:


packages/notifi-react-card/lib/hooks/useUnreadState.ts
```diff
@@ -22,6 +22,10 @@ export const useUnreadState = () => {
     () => (unreadNotificationCount > 0 ? true : false),
     [unreadNotificationCount],
   );
+
+  const updateNotificationCount = () => {
```
nimesh-notifi, line 26: considering this is setting to 0, I think it should be called clearNotificationCount

nimesh-notifi (approved):
minor fn name change, but lgtm

PR #567, opened 2024-05-14
new wallet support

ankush-notifi:
New Wallet Support - 
Rabby
Coinbase
Wallet Connect
Zerion
OKX
Rainbow
Binance


packages/notifi-wallet-provider/lib/utils/wagmi.ts
eric-notifi, line 1: I am wondering if it will be better rather than having this utils `wagmi.ts` in which we create the config with hardcoded `appName` and `walletConnect projectId`, we can expose this whole config out as an arg of `NotifiWalletProvider`. This way users will be able to input their own config.

ankush-notifi: I understand your point, and I appreciate your suggestion. However, I believe accepting the whole config as a prop might introduce complexity and increase the chances of errors if the config is not passed correctly.

Instead, considering our current setup and requirements, it seems we only need two pieces of information from the end user: the `app name` and the `project ID` for Wallet Connect. Simplifying it to these two parameters could make the integration smoother and reduce the risk of potential issues.
What do you think?

eric-notifi: Hey @ankush-notifi , are you going to have any changes into this PR (ex. exposing two more args ?)

eric-notifi (commented):
Do we still need this PR? If not shall we close it? @ankush-notifi

PR #585, opened 2024-05-20
Ankush/gmx fix

ankush-notifi:
Small bug fix


marukohao (approved):

marukohao:
It's better to make the disconnect button as a seperate component, but you can follow up in the following PR, let's merge first

PR #593, opened 2024-05-22
GMX integration

marukohao:


eric-notifi:
The cypress build failed in `notifi-react` I believe it should be caused by the conflict from [this PR](https://github.com/notifi-network/notifi-sdk-ts/pull/587/files). Might want to take a look can rebase 🙏.

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -30,14 +31,21 @@ export type TargetDocument = {
...
+  | 'phoneNumber'
+  | 'telegram'
+  | 'discord'
+  | 'slack'
+  | 'wallet';
 
 export type TargetInputFromValue = { value: string; error?: string };
 type TargetInputForm = Record<FormTarget, TargetInputFromValue>;
 
-export type TargetInputToggles = Record<ToggleTarget, boolean>;
```
eric-notifi, line 40: I think it is here, these section is mis-removed while doing rebase I believe

PR #595, opened 2024-05-23
reload issue fixed

ankush-notifi:
inbox page reload issue fixed - slide 7

### DIFF
--- a/packages/notifi-dapp-example/src/app/notifi/layout.tsx
+++ b/packages/notifi-dapp-example/src/app/notifi/layout.tsx
@@ -11,15 +11,18 @@ export default function NotifiSingupLayout({
 }: {
   children: React.ReactNode;
 }) {
-  const { wallets, selectedWallet } = useWallets();
+  const { wallets, selectedWallet, isAuthenticationVerified } = useWallets();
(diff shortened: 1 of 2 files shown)


packages/notifi-dapp-example/src/app/notifi/layout.tsx
```diff
@@ -11,15 +11,18 @@ export default function NotifiSingupLayout({
...
-  const { wallets, selectedWallet } = useWallets();
+  const { wallets, selectedWallet, isAuthenticationVerified } = useWallets();
   const pathname = usePathname();
   const router = useRouter();
 
   useEffect(() => {
-    if (!selectedWallet || !wallets[selectedWallet].walletKeys) {
+    if (
+      (!selectedWallet || !wallets[selectedWallet].walletKeys) &&
+      isAuthenticationVerified
```
marukohao, line 21: should we redirect to '/' when isAuthenticationVerified is false?

ankush-notifi: no, we should only redirect users to home '/'  page if `isAuthenticationVerified` status is true

PR #601, opened 2024-05-28
GMX integration

marukohao:


packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -80,18 +89,54 @@ const formatTelegramForSubscription = (telegramId: string) => {
 export type UpdateTargetInputs = <T extends 'form' | 'toggle'>(
   target: T extends 'form'
     ? Extract<Target, 'email' | 'phoneNumber' | 'telegram'>
-    : Extract<Target, 'discord' | 'slack'>,
-  value: T extends 'form' ? TargetInputFromValue : boolean,
+    : Extract<Target, 'discord' | 'slack' | 'wallet'>,
```
eric-notifi, line 92: I think we can use `FormTarget` and `ToggleTarget` types here

eric-notifi:
```tsx
? FormTarget : ToggleTarget
```

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -80,18 +89,54 @@ const formatTelegramForSubscription = (telegramId: string) => {
...
+  value: T extends 'form' ? { value: string; error?: string } : boolean,
```
eric-notifi, line 93: I guess using `TargetInputFromValue` instead of inline type definition is better

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -80,18 +89,54 @@ const formatTelegramForSubscription = (telegramId: string) => {
...
 ) => void;
 
+type FormTargetRenewArgs = {
+  target: Extract<Target, 'email' | 'telegram' | 'phoneNumber'>;
```
eric-notifi, line 97: Shall we use `FormTarget` here instead of inline type.

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -80,18 +89,54 @@ const formatTelegramForSubscription = (telegramId: string) => {
...
+  value: string;
+};
+
+type ToggleTargetRenewArgs = {
+  target: Extract<Target, 'discord' | 'slack' | 'wallet'>;
```
eric-notifi, line 102: Shall we use `ToggleTarget` here instead of inline type

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -80,18 +89,54 @@ const formatTelegramForSubscription = (telegramId: string) => {
...
+    args.target === 'discord'
+  );
+};
+
 export type NotifiTargetContextType = {
   isLoading: boolean;
   error: Error | null;
   updateTargetInputs: UpdateTargetInputs;
-  renewTargetGroup: () => Promise<void>;
+  renewTargetGroup: (args?: {
```
eric-notifi, line 132: Shall we  use `singleTargetRenewArgs` instead of `args` here?
1. It is more descriptive
2. consistent with the impl section

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -257,23 +317,62 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+
+          if (target === 'phoneNumber') {
+            formTarget = target;
+            formValue = isValidPhoneNumber(value) ? value : undefined;
+          }
+
+          data = {
+            ...targetGroupToBeSaved,
+            [formTarget]: formValue,
+          };
```
eric-notifi, line 349: Just being curious about this section.
As far as I know, this block only happens when we want to update `singleTarget`. But I think others will get updated if we use destruct `targetGroupToBeSaved` in which there might have some changed target?
Or do I miss anything?

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -548,9 +653,49 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+  const refreshWeb3Target = useCallback(
+    async (web3Target?: Types.Web3TargetFragmentFragment) => {
+      if (web3Target) {
+        switch (web3Target.isConfirmed) {
+          case false:
+            updateTargetInfoPrompt('wallet', {
+              type: 'cta',
+              message: 'Sign Wallet',
+              onClick: () => {
+                // TODO: add action
```
eric-notifi, line 665: Shall we add `action` here?

packages/notifi-react/utils/topic.ts
eric-notifi, line 1: Hmmmm .... not sure why this `topic.ts` become a new file even we have this one in `main`?? 
Also this one seems outdated, which does not adopt the newest change for migrating alertName from topicName to fusionEventTypeId??

packages/notifi-dapp-example/src/hooks/useNotifiTargetListener.ts
eric-notifi, line 1: Not sure if we still need this file as we have already add this feature in `NotifiTargetContext`

packages/notifi-dapp-example/src/utils/topic.ts
eric-notifi, line 1: Ah ... I think this `topic.ts` is the one should be located in `notifi-react`

packages/notifi-dapp-example/public/logos/CB1.png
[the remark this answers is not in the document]
marukohao: @ankush-notifi Could you address this one?

packages/notifi-react/lib/context/NotifiTargetContext.tsx
marukohao: @ankush-notifi  can you address this one?

packages/notifi-react/lib/context/NotifiTargetContext.tsx
marukohao: @ankush-notifi FYI

ankush-notifi: You're correct, Eric. However, if we don't  pass in `targetGroupToBeSaved` directly, the frontendClient.ensureTargetGroup function will remove existing target groups.

eric-notifi: Ah ... what about using the data from `targetData` instead of `targetGroupToBeSaved` as `targetData` reflects the data stored in BE. wdyt?

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -658,7 +658,9 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
               type: 'cta',
               message: 'Sign Wallet',
               onClick: () => {
-                // TODO: add action
+                // Note: -
+                // No action required here as this is only possible on the client side
+                // based on the connected wallet instance. We're handling this on the frontend.
```
eric-notifi, line 663: In my opinion, we can migrate those xmtp sign methods to `NotifiTargetContext`. this way:
1. we can simplify the component logic making it easier to follow.
2. we can reuse that method possible in the future if other components need that.

eric-notifi (approved):
LGTM!!
May need to take a look at the conflict before having this merged.🙏

eric-notifi (changes requested):
Ah.... Looks like the [last merged PR](https://github.com/notifi-network/notifi-sdk-ts/pull/608) is missing. Could you double check if we did pull before force-merge ??

eric-notifi (approved):
hmmm NVM .... I think it is still included. Wried that we cannot see it in commit history...
🫨

PR #608, opened 2024-06-04
Ankush/xmtp refactor

ankush-notifi:
Moved `Sign Wallet - XMTP` logic into `notifi-react` package


packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -661,6 +664,156 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+  }, [
+    walletWithSignParams,
+    targetData,
+    signCoinbaseSignature,
+    frontendClient,
+    xmtp,
+    getSignature,
+  ]);
+
+  const signWallet = useCallback(async () => {
```
eric-notifi, line 813: Just being curious, do we have to have the extra layer for this `signWallet`?

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -661,6 +664,156 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
 
+  const getSignature = useCallback(
+    async (message: ArrayLike<number> | string) => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      const signature = await walletWithSignParams.signMessage(message);
+      let hexString = '0x';
+
+      for (let i = 0; i < Object.keys(signature).length; i++) {
+        // Convert each value to a two-digit hexadecimal string and append it to the result
```
eric-notifi, line 675: Not sure if this `getSignature` is going to be an general method that will adopt different wallet? or only for coinbase.
Because we supports multi-chain, the signature might vary. Not sure if you have an idea how to handle those case?

eric-notifi (commented):
Generally LGTM with some nits below.
And also curious about the relationship between different methods.
1. getSignature: it looks like just a method converting the signature signed by `walletWithSignParams.signMessage(message)`. I am not sure if we can simply have a util method doing such conversion? something like `reformateSignatureForWalletTarget` 
2. `signCoinbaseSignature` is called in `xmtpXip42Impl`, does it mean `xmtpXip42Impl` is specifically for coinbase wallet?
CC: @kai-notifi @ankush-notifi

packages/notifi-react/lib/context/NotifiTargetContext.tsx
ankush-notifi: yes, I think so, if we need additional step

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -669,14 +673,7 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
       const signature = await walletWithSignParams.signMessage(message);
-      let hexString = '0x';
-
-      for (let i = 0; i < Object.keys(signature).length; i++) {
-        // Convert each value to a two-digit hexadecimal string and append it to the result
-        hexString += signature[i].toString(16).padStart(2, '0');
-      }
-
-      return hexString;
+      return reformatSignatureForWalletTarget(signature);
```
eric-notifi, line 676: I think we need to get rid of the type issue for scalability and maintainability as it is SDK.
I think this type error is from mis handling chain specific differences. We might want to handle those case, just an conceptual example:
```ts
switch (walletWithSignParams.walletBlockchain) {
        case 'AVALANCHE':
        case 'ETHEREUM':
        case ...etc:
        // handle get signature
        break;
        default:
        // handle error
      }
```

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -661,6 +668,151 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+      // TODO: Add logic for rest of the chains
+      switch (walletWithSignParams.walletBlockchain) {
+        case 'AVALANCHE':
+        case 'ETHEREUM':
+          signature = await walletWithSignParams.signMessage(
+            message as Uint8Array,
+          );
+          break;
+        default:
+          signature = '';
```
eric-notifi, line 684: Shall we handle an error here ex. setError('ERROR: not supported chain') just in case.

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -661,6 +668,151 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+  const getSignature = useCallback(
+    async (message: Uint8Array | string) => {
+      let signature: Uint8Array | string = '';
+
+      // TODO: Add logic for rest of the chains
+      switch (walletWithSignParams.walletBlockchain) {
+        case 'AVALANCHE':
+        case 'ETHEREUM':
+          signature = await walletWithSignParams.signMessage(
+            message as Uint8Array,
```
eric-notifi, line 680: we cast the message as `Uint8Array`, but we actually pass in `string` type, would it possibly cause issue.
Or we want to convert to actually Uint8Arry before passing in `signMessage`

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -661,6 +668,151 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+    };
+
+    const client = await xmtp.initialize({ options, signer });
+
+    if (client === undefined) {
+      throw Error('XMTP client is uninitialized. Please try again.');
+    }
+
+    // TODO: get senderAddress from target
+    const senderAddress = '0xE80E42B5308d5b137FC137302d571B56907c3003';
```
eric-notifi, line 749: Just wondering what is this address? does this one represent `notifi`?? or Tenant specific

eric-notifi (commented):
Awoseme, the structure looks much clear to me now. Thank you.
Just some final nits left. please kindly take a look.🙏

packages/notifi-react/lib/context/NotifiTargetContext.tsx
ankush-notifi: @kai-notifi, I think it's more of your question.

kai-notifi: This is tenant specific just for gmx. In the near future we should take this from the web3 target. I'm currently working on that so we can support that

eric-notifi: After disucssing with Kai, the sender address will be eventually retrieve from `Web3Target` ( `targetData.wallet.senderAddress` ). 
Just for VIZ CC: @ankush-notifi @kai-notifi

eric-notifi: It means this feature will not be able to use for other clients except GMX. so we might want to add a comment to make it clear 🙏
Others LGTM . Thanks for the great effort @ankush-notifi

eric-notifi (approved):

packages/notifi-react/lib/context/NotifiTargetContext.tsx
eric-notifi: Feel free to merge after adding the comments for this section. And please also check it can be built and also functionally working on GMX. Thanks again! ✦ @ankush-notifi

PR #614, opened 2024-06-10
fix demo preview render issue

marukohao:
preview react card doesn't render correctly with different demoPreview value that be passed in, like 'history', 'signup' or 'preview' in AP

add 'demoPreview' as dependecy

https://notifi.atlassian.net/jira/software/c/projects/MVP/boards/1?assignee=62957038658d580068882428&selectedIssue=MVP-4267

### DIFF
--- a/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
+++ b/packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
@@ -191,7 +191,7 @@ export const SubscriptionCardV1: React.FC<SubscriptionCardV1Props> = ({
 
       return { state: 'history' };
     });
-  }, []);
+  }, [demoPreview]);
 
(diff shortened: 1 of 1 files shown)


packages/notifi-react-card/lib/components/subscription/SubscriptionCardV1.tsx
```diff
@@ -191,7 +191,7 @@ export const SubscriptionCardV1: React.FC<SubscriptionCardV1Props> = ({
 
       return { state: 'history' };
     });
-  }, []);
+  }, [demoPreview]);
```
eric-notifi, line 194: This `useEffect` section is supposed to only render once when the card is mounted. It is a potential risk to add the `demoPreview` into dependency to make this view re-rendered.
Do you think we can make the card rerendered in upper level in DAPP (AP) instead of in card level.

PR #616, opened 2024-06-12
notifi-dapp-example (ODF specific): make sure localStorage is removed before route back to root

eric-notifi:
As title, Try to resolve the edge case the the infinite spin showing up after re-deploying

No need
N/A

### DIFF
--- a/packages/notifi-dapp-example/src/hooks/useNotifiRouter.ts
+++ b/packages/notifi-dapp-example/src/hooks/useNotifiRouter.ts
@@ -53,12 +53,16 @@ export const useNotifiRouter = () => {
 
   useEffect(() => {
     if (loginError) {
-      handleRoute('/');
       if (selectedWallet) {
         wallets[selectedWallet].disconnect();
       } else if (injectiveSelectedWallet) {
         injectiveWallets[injectiveSelectedWallet].disconnect();
       }
+      setTimeout(() => {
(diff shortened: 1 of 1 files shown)


packages/notifi-dapp-example/src/hooks/useNotifiRouter.ts
```diff
@@ -53,12 +53,16 @@ export const useNotifiRouter = () => {
...
       if (selectedWallet) {
         wallets[selectedWallet].disconnect();
       } else if (injectiveSelectedWallet) {
         injectiveWallets[injectiveSelectedWallet].disconnect();
       }
+      setTimeout(() => {
+        /** NOTE: preserve 0.5s to make sure cache is cleaned before route back to root dir*/
+        console.log(`LOGIN ERROR: ${JSON.stringify(loginError)}`);
+        handleRoute('/');
+      }, 5000);
```
ankush-notifi, line 65: I noticed that the note mentions a 0.5-second delay to ensure the cache is cleaned before routing back to the root directory. However, the code currently has a 5-second delay. Could you please confirm if this is intentional or if it should be 500 milliseconds instead of 5000 milliseconds as per the note?

cc @eric-notifi

eric-notifi: Nice catch!! I was gona do 500. Thanks @ankush-notifi.

PR #623, opened 2024-06-24
xion-wallet sdk support

ankush-notifi:
xion-wallet sdk support
https://notifi.atlassian.net/browse/MVP-4964
https://notifi.atlassian.net/browse/MVP-4965

### DIFF
--- a/packages/notifi-dapp-example/src/components/WalletSelectModal.tsx
+++ b/packages/notifi-dapp-example/src/components/WalletSelectModal.tsx
@@ -14,16 +14,7 @@ export const WalletSelectModal: FC<WalletSelectModalProps> = ({
 }) => {
   const { wallets } = useWallets();
   const availableWallets = objectKeys(wallets);
-  let walletsRequiredForGMX: typeof availableWallets = [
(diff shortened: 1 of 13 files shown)


packages/notifi-dapp-example/src/components/WalletSelectModal.tsx
```diff
@@ -14,16 +14,7 @@ export const WalletSelectModal: FC<WalletSelectModalProps> = ({
...
-    'metamask',
-    'coinbase',
-    'rabby',
-    'walletconnect',
-    'binance',
-    'okx',
-    'rainbow',
-    'zerion',
-  ];
+  let walletsRequiredForGMX: typeof availableWallets = ['xion'];
```
eric-notifi, line 17: Ah ... well, I understand this one is specifically for Xion dapp. 
However, this is good. But do you think it might be better to separate this PR into two:
1. for the generic sections which can be be use commonly in the future. (ex. notifi-wallet-provider) --> against main branch 
2. for others, we can make it specifically for Xion

PR #624, opened 2024-06-24
Style Xion Dapp 

marukohao:


eric-notifi (commented):
Well, Would it be a better idea to create a folk version repo. There will be few benifits.
1. Clients related codebase will not always be exposed publicly.
2. The branch will not be growing big rapidly, which might confuse users.

> **NOTE**
> I think only those changes which possibly need to be merged into main would be the branch of notifi-sdk-ts. How do you think?
CC: @marukohao @ankush-notifi @nimesh-notifi

PR #625, opened 2024-06-25
Ankush/xion wallet

ankush-notifi:
xion-wallet sdk support
https://notifi.atlassian.net/browse/MVP-4964

### DIFF
--- a/packages/notifi-wallet-provider/lib/types.ts
+++ b/packages/notifi-wallet-provider/lib/types.ts
@@ -20,17 +20,27 @@ export type WalletKeysBase = {
   bech32: string;
   base64: string;
   hex: string;
+  granter: string;
+  grantee: string;
+  pubKey: string;
 };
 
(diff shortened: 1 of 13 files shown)


packages/notifi-wallet-provider/lib/types.ts
```diff
@@ -20,17 +20,25 @@ export type WalletKeysBase = {
   bech32: string;
   base64: string;
   hex: string;
+  grantee: string;
```
eric-notifi, line 23: I think we still need to know about the format about the grantee so that we can reuse this field for those upcoming chains potentially.
So, If we can not get to know it now, we might want to put a todo item, something like below.
`//TODO: specifically for Xion now, make it generic pattern, checking the format`

eric-notifi (approved):
Generally LGTM, please check the nit before merging. Thx!

PR #640, opened 2024-07-09
Log Out on Auth Failure

mayuranganesathas:


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -361,6 +361,7 @@ export class NotifiFrontendClient {
           authorization = newAuthorization;
         }
       } catch (e: unknown) {
+        await this.logOut();
```
hans-notifi, line 364: I would try to parse the error and see if it's an `AUTH_NOT_AUTHORIZED` error code. We shouldn't be logging out if the API server is unreachable or if an unexpected exception occurs, for example.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -361,6 +361,7 @@ export class NotifiFrontendClient {
           authorization = newAuthorization;
         }
       } catch (e: unknown) {
```
eric-notifi, line 363: As far as I know, This section indicate that the token expired date will be extended while being close to the expired date of jwt. 
A question: wondering if this method is already deprecated and will ALWAYS fail, meaning it will 100% goes into `catch` block? if this is the case, why don't we just simply remove all this `refreshAuthoriztion` related logic (line 353 - 375) and also add some handling logic instead of making it bening  logged out eariler?

hans-notifi: The `refreshAuthorization` method still works perfectly well as of now. It's marked deprecated because we want to slowly move the SDK over to the Initial/Refresh/Normal/Elevated token pattern that I implemented as part of 2FA support in Admin Portal. Once we change _all_ clients to use the new pattern, we can safely delete `refreshAuthorization`.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: will add a fix!

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: Ohoh, I see. Then it makes sense. thank you!

eric-notifi (commented):
Just a small nit. Please rebase main so that Vercel deployment build can run through
```bash
git rebase main
# and then force push to branch again
```
CC: @mayuranganesathas

PR #643, opened 2024-07-11
MVP-5118 - Fix invalid TG & Email error disappearing after the window…

kaan-notifi:
Fix email and Telegram input validation and error persistence

This PR addresses issues with email and Telegram input validation and error persistence in the NotifiTargetContext. The changes improve the user experience by clearing errors for empty inputs and maintaining validation state on refresh and window refocus in between switching tabs.

Key changes:
- Add email and Telegram validation functions with proper handling of empty inputs
- Update `updateTargetInputs` function to apply validation for email and Telegram inputs
- Modify `refreshTargetDocument` function to preserve validation state on refresh for both email and Telegram
- Update `targetGroupToBeSaved` object to use validated email and Telegram inputs
- Fix TypeScript errors related to spreading non-object types

These updates ensure that:
1. Error messages disappear when inputs are cleared
2. Validation states persist when the target document is refreshed
3. The "Next" button remains disabled for invalid inputs after a refresh

Testing:
- Verified that email and Telegram error messages clear when inputs are emptied
- Confirmed that validation states persist after refreshing the target document
- Ensured that the "Next" button behavior is consistent with input validity

### DIFF
--- a/packages/notifi-react/lib/context/NotifiTargetContext.tsx
+++ b/packages/notifi-react/lib/context/NotifiTargetContext.tsx
@@ -45,7 +45,11 @@ export type Target =
 export type FormTarget = Extract<Target, 'email' | 'phoneNumber' | 'telegram'>;
 export type ToggleTarget = Extract<Target, 'discord' | 'slack' | 'wallet'>;
 
-export type TargetInputFromValue = { value: string; error?: string };
+export type TargetInputFromValue = {
+  value: string;
+  error?: string;
+  isValid?: boolean;
+};
 type TargetInputForm = Record<FormTarget, TargetInputFromValue>;
 
 export type TargetInputToggles = Record<ToggleTarget, boolean>;
@@ -198,12 +202,14 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
   const targetGroupToBeSaved: TargetGroupInput = {
     name: 'Default',
     emailAddress:
-      targetInputs.email.value === '' ? undefined : targetInputs.email.value,
+      targetInputs.email.value === '' || !targetInputs.email.isValid
+        ? undefined
+        : targetInputs.email.value,
     phoneNumber: isValidPhoneNumber(targetInputs.phoneNumber.value)
       ? targetInputs.phoneNumber.value
       : undefined,
     telegramId:
-      targetInputs.telegram.value === ''
+      targetInputs.telegram.value === '' || !targetInputs.telegram.isValid
         ? undefined
         : formatTelegramForSubscription(targetInputs.telegram.value),
     discordId: targetInputs.discord ? 'Default' : undefined,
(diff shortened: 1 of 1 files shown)


packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -310,16 +316,74 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+      isValid,
+      error: isValid ? undefined : 'Invalid telegram ID',
+    };
+  };
+
   const updateTargetInputs = useCallback(
     <T extends 'form' | 'toggle'>(
       target: T extends 'form' ? FormTarget : ToggleTarget,
-      value: T extends 'form' ? { value: string; error?: string } : boolean,
+      value: T extends 'form' ? TargetInputFromValue : boolean,
```
eric-notifi, line 352: This is also included in the incoming stage branch which will be merged into main next Tuesday.

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -310,16 +316,74 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+    if (email === '') {
+      return { isValid: true };
+    }
+    const isValid = EMAIL_REGEX.test(email);
+    return {
+      isValid,
+      error: isValid ? undefined : 'Invalid email address',
+    };
+  };
+
```
eric-notifi, line 335: In context, we might want to simply just give the error message if there is. 
And in component, we can decide the component logic more flexible using the util method mentioned ablve

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -310,16 +316,74 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+            // For toggle inputs
+            return {
+              ...prev,
+              [target]: value,
+            };
+          }
+          // If we reach here, it means we got an unexpected input
+          console.error(`Unexpected input for target ${target}:`, value);
+          return prev; // Return previous state unchanged
+        });
```
eric-notifi, line 386: The same comment as above.  I would not put this logic in Context. Context should be subjective. 
Defining the valid or invalid should be outside of context.
See the example [here](https://github.com/notifi-network/notifi-sdk-ts/blob/79b728739e350e118da59a3305f05e9cef1dc419/packages/notifi-react/lib/components/TargetInputs.tsx#L63)

packages/notifi-react/lib/context/NotifiTargetContext.tsx
```diff
@@ -423,27 +487,47 @@ export const NotifiTargetContextProvider: FC<PropsWithChildren> = ({
...
+          (it) => it?.name === 'Default',
+        ),
+        slack: !!targetGroup?.slackChannelTargets?.find(
+          (it) => it?.name === 'Default',
+        ),
+        wallet: !!targetGroup?.web3Targets?.find(
+          (it) => it?.name === 'Default',
+        ),
+      };
+    });
```
eric-notifi, line 531: Same comment as above

eric-notifi (commented):
Generally speaking. The design pattern of the context is to give the "Objective" data to those developer who might want to consume the context in their components or Dapp. So we do not predefine the `valid` or `invlalid` which is subjective. Developers can define that in components or dapp level by setting setting the `error` property state using `updateTargetInput`
The input field focus and blur behavior should be addressed in component level instead of  Context level.
Here is the example :https://github.com/notifi-network/notifi-sdk-ts/blob/79b728739e350e118da59a3305f05e9cef1dc419/packages/notifi-react/lib/components/TargetInputs.tsx#L63

eric-notifi:
@kaan-notifi , Shall we close this PR if no longer need?

PR #654, opened 2024-07-22
Support net fee alert - add group topic support to topic stack row

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
+++ b/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
@@ -119,9 +119,11 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
             </div>
           )}
           <div
-            className={`absolute top-12 left-0 w-[247px] ${
-              getSubscriptionValueOrRef() === 'pricePairs' ? '' : 'pt-[36px]'
-            } bg-notifi-card-bg rounded-lg shadow-lg z-10 max-h-[180px] overflow-y-auto mt-1`}
+            className={`absolute left-0 w-[247px] ${
+              getSubscriptionValueOrRef() === 'pricePairs'
+                ? 'top-12 rounded-lg'
+                : 'top-[86px] rounded-b-lg'
+            } bg-notifi-card-bg shadow-lg z-10 max-h-[140px] overflow-y-auto mt-1`}
           >
             {filteredOptions.map((option) => (
               <div
--- a/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
+++ b/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
@@ -50,21 +50,15 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
         className="rounded-md bg-notifi-card-bg text-start text-notifi-text"
         onClick={() => setIsDropdownOpen((prev) => !prev)}
       >
-        {props.subscriptionValue
-          ? props.subscriptionValue.label
-          : (
-              <div className="text-notifi-text-light">
-                {getSubscriptionValueOrRef() === 'pricePairs'
-                  ? 'Select Pair'
-                  : 'Select Market'}
-              </div>
-            ) ?? (
-              <div className="text-notifi-text-light">
-                {getSubscriptionValueOrRef() === 'pricePairs'
-                  ? 'Select Pair'
-                  : 'Select Market'}
-              </div>
-            )}
+        {props.subscriptionValue ? (
+          props.subscriptionValue.label
+        ) : (
+          <div className="text-notifi-text-light">
+            {getSubscriptionValueOrRef() === 'pricePairs'
+              ? 'Select Pair'
+              : 'Select Market'}
+          </div>
+        )}
       </div>
       {isDropdownOpen && (
         <div>
--- a/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
+++ b/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
@@ -19,11 +19,29 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
   const { inputs } = useNotifiTenantConfigContext();
   const subscriptionValueOrRef = props.subscriptionValueRef;
   const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+  const [isTermSelectDropdownOpen, setIsTermSelectDropdownOpen] =
+    React.useState(false);
+  const [selectTerm, setSelectTerm] = React.useState<string>('All');
   const subscriptionValueOptions = resolveObjectArrayRef(
     'subscriptionValueOptions',
     subscriptionValueOrRef,
     inputs,
   );
+
+  const getSubscriptionValueOrRef = () => {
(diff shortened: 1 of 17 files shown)


packages/notifi-dapp-example/src/components/TopicStack.tsx
```diff
@@ -12,46 +12,65 @@ import {
...
     if (!input) return;
     return Object.values(input)[0];
   }, [getAlertFilterOptions]);
 
-  const fusionEventMetadata = getFusionEventMetadata(props.topic);
+  const { isLoading } = useNotifiTopicContext();
+  const resolved = resolveAlertName(benchmarkAlert.alertName);
+  const fusionTopic = getFusionTopic(resolved.fusionEventTypeId);
+  if (!fusionTopic) return null;
+  const fusionEventMetadata = getFusionEventMetadata(fusionTopic);
```
eric-notifi, line 47: As mentioned before, If we do not group the alerts with the same filterOptions, it will result in rendering multiple duplicate alerts. 
Not sure if I am missing anything?

marukohao: Yes we have the logic to filter out the duplicate alerts(with same filterOptions) in TopicStackRow

packages/notifi-dapp-example/src/components/TopicStackRow.tsx
```diff
@@ -1,34 +1,73 @@
...
-    getFusionEventMetadata(props.topic)?.uiConfigOverride?.topicDisplayName ??
-    '';
-
+    getFusionEventMetadata(benchmarkTopic)?.uiConfigOverride
+      ?.topicDisplayName ?? '';
+  const isTradingPairAlert = title.includes('Pair');
   const [isTopicStackRowInputVisible, setIsTopicStackRowInputVisible] =
     React.useState(topicStackAlerts.length > 0 ? false : true);
 
+  const alertsStack = React.useMemo(() => {
```
marukohao, line 38: We have filter out duplicate alerts with same filterOptions and made sure we have the duplicate alerts only render once here @eric-notifi

marukohao:
How should I fix the Vercel deployment fail? @eric-notifi

packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
```diff
@@ -34,26 +52,112 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
...
+                {getSubscriptionValueOrRef() === 'pricePairs'
+                  ? 'Select Pair'
+                  : 'Select Market'}
+              </div>
+            ) ?? (
+              <div className="text-notifi-text-light">
+                {getSubscriptionValueOrRef() === 'pricePairs'
+                  ? 'Select Pair'
+                  : 'Select Market'}
+              </div>
```
eric-notifi, line 66: Looks like these two statements are the same. Do I miss anything? 
And just my opinion, using such kind of condition rendering for label empty state is not scalable.
Do you think we can just extract the value from the ref? like
```ts
function splitCamelCaseToString(input) {
  return input
    .replace(/([a-z])([A-Z])/g, '$1 $2')
    .toLowerCase();
}
 
 <div>Select {splitCamelCaseToString(getSubscriptionValueOrRef())} </div>

File: packages/notifi-dapp-example/src/components/TopicStackRow.tsx [hunk 4]
```diff
@@ -94,7 +135,7 @@ export const TopicStackRow: React.FC<TopicStackRowProps> = (props) => {
               setIsTopicStackRowInputVisible(true);
             }}
           >
-            + Add Pair
+            {isTradingPairAlert ? '+ Add Pair' : '+ Add Fee Rate'}
```
eric-notifi, line 138: Similar comment as above, an approach is to use regex to resolve the `ref` and rendered the resolved value.

packages/notifi-dapp-example/src/components/TopicStackRowInput.tsx
```diff
@@ -177,36 +206,67 @@ export const TopicStackRowInput: React.FC<TopicStackRowInputProps> = (
...
+                            );
+                          setFilterOptionsToBeSubscribed(
+                            updatedAlertFilterOptiopns,
+                          );
+                        },
+                      }}
+                    />
+                  );
+                },
+              )}
```
eric-notifi, line 269: Just being curious, do we still need to ensure the correct order of userInputParms. Not sure if it will be confusing for user as the order is controlled by AP.

packages/notifi-react/lib/context/NotifiTopicContext.tsx
```diff
@@ -30,16 +29,18 @@ import {
 import { useNotifiFrontendClientContext } from './NotifiFrontendClientContext';
 import { useNotifiTenantConfigContext } from './NotifiTenantConfigContext';
 
+export type TopicWithFilterOption = {
+  topic: FusionEventTopic;
+  filterOptions: FusionFilterOptions;
+  subscriptionValue?: string;
+  customAlertName?: string;
+};
+
```
eric-notifi, line 38: Ah ...... Shall we merged MVP-6153. And these can be imported instead of use the redundant code.

eric-notifi (commented):
Just some nit for making the example more scalable and generic so that it can also be merged into main.
Also, Let me finish up the MVP-5163 and merge into main so that we will not have copy-past redundant code in this PR.

eric-notifi:
> How should I fix the Vercel deployment fail? @eric-notifi

Ohoh, the vercel deploy is particullarly for `notifi-react-exaple-v2` repo. So it wont affect this PR. But regardless, let me redeploy it.

packages/notifi-dapp-example/src/components/TopicStackRowInput.tsx
marukohao: Yes, I think we should for GMX as for the initial topics we haven't update the userInputParams with the correct order, but for new create topics, it should be the correct order, just incase we don't mess up the rendering for the  existing alerts

packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
marukohao: Ah you are right, I need to remove one of the statement
The label is customized for different topics right now, so we can not extract the value from ref, I think in the future we may need to include a new field in metadata to include the select input placeholder.

eric-notifi: Yeah, I understand the label is customized. But what I am thinking is that if they set the ref in AP is `pricePair`, the empty label showing on the UI would be `Select Price Pair`. so it will separate the camel case to a normal phase. 
So it can be a default behavior which is controllable through AP using the `input ref`

marukohao: Ah you mean we are doing something like this?

```
function splitCamelCaseToString(input) {
  return input
    .replace(/([a-z])([A-Z])/g, '$1 $2')
    .toLowerCase();
}
( <div className="text-notifi-text-light">
      {getSubscriptionValueOrRef() === 'pricePairs'
             ? 'Select Pair'
             : 'Select Market'}
 </div>
) ?? (
  <div>Select {splitCamelCaseToString(getSubscriptionValueOrRef())} </div>)
```

PR #658, opened 2024-07-25
Add prefix/suffix and group topic support to topic stack row in dapp example

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/next.config.mjs
+++ b/packages/notifi-dapp-example/next.config.mjs
@@ -2,29 +2,30 @@
 import CopyPlugin from "copy-webpack-plugin";
 
 const nextConfig = {
-  output: 'export' /** BUILD ONLY */,
+  // output: 'export' /** BUILD ONLY */,
   images: {
     formats: ['image/webp'],
   },
-  typescript: {
-    // !! WARN !!
-    // Dangerously allow production builds to successfully complete even if
-    // your project has type errors.
-    // !! WARN !!
-    ignoreBuildErrors: true,
-  },
-  eslint: {
-    // Warning: This allows production builds to successfully complete even if
-    // your project has ESLint errors.
-    ignoreDuringBuilds: true,
-  },
+  // typescript: { /** BUILD ONLY */
+  //   // !! WARN !!
+  //   // Dangerously allow production builds to successfully complete even if
+  //   // your project has type errors.
+  //   // !! WARN !!
+  //   ignoreBuildErrors: true,
+  // },
+  // eslint: { /** BUILD ONLY */
+  //   // Warning: This allows production builds to successfully complete even if
+  //   // your project has ESLint errors.
+  //   ignoreDuringBuilds: true,
+  // },
   webpack: (
     config,
     { buildId, dev, isServer, defaultLoaders, nextRuntime, webpack }
   ) => {
     config.plugins.push(new CopyPlugin({
       patterns: [
-        { from: "public/wasm", to: "./server/chunks" },
+        // { from: "public/wasm", to: "./server/chunks" }, /** BUILD ONLY */
+        { from: "public/wasm", to: "./server/vendor-chunks/" },
       ],
     }))
     return config;
--- a/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
+++ b/packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
@@ -19,11 +19,29 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
   const { inputs } = useNotifiTenantConfigContext();
   const subscriptionValueOrRef = props.subscriptionValueRef;
   const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+  const [isTermSelectDropdownOpen, setIsTermSelectDropdownOpen] =
+    React.useState(false);
+  const [selectTerm, setSelectTerm] = React.useState<string>('All');
   const subscriptionValueOptions = resolveObjectArrayRef(
     'subscriptionValueOptions',
     subscriptionValueOrRef,
     inputs,
   );
+
+  const getSubscriptionValueOrRef = () => {
+    if (subscriptionValueOrRef.type === 'ref') {
+      return subscriptionValueOrRef.ref;
+    } else {
+      return subscriptionValueOrRef.value;
+    }
+  };
+
+  const filteredOptions = subscriptionValueOptions.filter((option) => {
+    if (selectTerm === 'All') {
+      return true;
+    }
+    return option.label.startsWith(selectTerm);
+  });
 
   return (
     <div className="relative ml-14 w-[247px] h-12 bg-notifi-card-bg rounded-md mb-3 text-notifi-text p-3 border-0 focus:border-0 focus-visible:border-0 cursor-pointer">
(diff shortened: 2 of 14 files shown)


packages/notifi-dapp-example/next.config.mjs
```diff
@@ -2,29 +2,30 @@
...
+  //   // Dangerously allow production builds to successfully complete even if
+  //   // your project has type errors.
+  //   // !! WARN !!
+  //   ignoreBuildErrors: true,
+  // },
+  // eslint: { /** BUILD ONLY */
+  //   // Warning: This allows production builds to successfully complete even if
+  //   // your project has ESLint errors.
+  //   ignoreDuringBuilds: true,
+  // },
```
eric-notifi, line 20: Rather than using comment, we can dynamic define the build or dev. Here is the example, hope it helps: https://github.com/notifi-network/notifi-sdk-ts/blob/main/packages/notifi-react-example-v2/next.config.mjs

packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
```diff
@@ -19,41 +19,139 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
   const { inputs } = useNotifiTenantConfigContext();
   const subscriptionValueOrRef = props.subscriptionValueRef;
   const [isDropdownOpen, setIsDropdownOpen] = React.useState(false);
+  const [isTermSelectDropdownOpen, setIsTermSelectDropdownOpen] =
+    React.useState(false);
+  const [selectTerm, setSelectTerm] = React.useState<string>('All');
```
eric-notifi, line 24: just a super small nit, `selectTerm` looks like a method. How do you think?

packages/notifi-dapp-example/src/components/SubscriptionValueInput.tsx
```diff
@@ -19,41 +19,139 @@ export const SubscriptionValueInput: React.FC<SubscriptionValueInputProps> = (
...
-          : <div className="text-notifi-text-light">Select Pair</div> ?? (
-              <div className="text-notifi-text-light">Select Pair</div>
-            )}
+        {props.subscriptionValue ? (
+          props.subscriptionValue.label
+        ) : (
+          <div className="text-notifi-text-light">
+            {getSubscriptionValueOrRef() === 'pricePairs'
+              ? 'Select Pair'
+              : 'Select Market'}
```
eric-notifi, line 59: NOTE: need help to open the ticket to make it dynamic.  thanks!!

packages/notifi-dapp-example/src/components/TopicOptions.tsx
```diff
@@ -213,31 +254,57 @@ export const TopicOptions = <T extends TopicRowCategory>(
...
+                  evt.target.value,
+                  props.userInputParam.customInputConstraints,
+                );
+                setCustomInput(value.toString());
+                selectOrInputValue(value);
               }}
-              placeholder={props.placeholder ?? 'Custom'}
+              placeholder={
+                props.userInputParam.kind === 'price'
+                  ? 'Enter Price'
```
eric-notifi, line 279: Hmmm .... I guess we will no longer have price kind. Shall we make a more natural term to fallback to?

packages/notifi-dapp-example/src/components/TopicOptions.tsx
```diff
@@ -183,19 +210,33 @@ export const TopicOptions = <T extends TopicRowCategory>(
...
+                  } else {
+                    if (
+                      isUserInputValid(
+                        props.userInputParam.kind,
+                        evt.target.value,
+                        props.userInputParam.customInputConstraints,
+                      )
+                    ) {
+                      selectOrInputValue(evt.target.value);
+                    }
```
eric-notifi, line 231: Not sure if we still need the following hardcode section:

```ts
if (props.userInputParam.options[0]) {
                  if (
                    Number(evt.target.value) > 99 ||
                    (Number(evt.target.value) < 1 && evt.target.value !== '')
                  ) {
                    setErrorMessage('Please enter a value between 1 and 99');
                  }
                } else {
```

packages/notifi-dapp-example/src/components/TopicOptions.tsx
```diff
@@ -213,31 +254,57 @@ export const TopicOptions = <T extends TopicRowCategory>(
...
+              customInput ? (
+                <div className="absolute right-4 bottom-3">
+                  {'suffix' in prefixAndSuffix && prefixAndSuffix.suffix}
+                </div>
+              ) : null
+            ) : (
               <div className="absolute right-3 bottom-3">
-                {'prefix' in prefixAndSuffix && prefixAndSuffix.prefix}
+                {'suffix' in prefixAndSuffix && prefixAndSuffix.suffix}
               </div>
```
eric-notifi, line 306: Does this two conditions duplicated?

packages/notifi-dapp-example/src/components/TopicRow.tsx
```diff
@@ -89,6 +93,17 @@ export const TopicRow = <T extends TopicRowCategory>(
...
+    userInputParams: UserInputParam<UiType>[],
+  ) => {
+    if (userInputParams.length > 0 && userInputParams[0].uiType === 'radio') {
+      return userInputParams;
+    } else {
+      const reversedParams = [...userInputParams].reverse();
+      return reversedParams;
+    }
+  };
+
```
eric-notifi, line 106: NOTE: per talk, we might also want add a todo or a ticket so that we do not forget to remove this section

marukohao: https://notifi.atlassian.net/browse/MVP-5238 I have the ticket created, feel free to edit if there are some thing that I missed

packages/notifi-dapp-example/src/components/TopicOptions.tsx
marukohao: We'll still have price kind for trading pair alert

packages/notifi-dapp-example/src/components/TopicOptions.tsx
marukohao: I think we do need the hardcode section, as there is no uppper/lower bound set up for liquidation alert

packages/notifi-dapp-example/src/components/TopicOptions.tsx
marukohao: nope, one for with Threshold Option button, one for without Threshold Option button.

packages/notifi-dapp-example/src/components/TopicOptions.tsx
eric-notifi: Hmmmm... I think we will eventually remove the condition `props.userInputParam.options[0]`. If it is not proper to do it now, shall we make a `TODO:` comment so with can easily understand this part and upcoming action.

packages/notifi-dapp-example/src/components/TopicOptions.tsx
eric-notifi: I think we will eventually remove the condition `props.userInputParam.options[0]`. If it is not proper to do it now, shall we make a `TODO:` comment so with can easily understand this part and upcoming action.

eric-notifi (approved):
Just a final nit, I think we would eventually remove the condition `props.userInputParam.options[0]`. If it is not proper to do it now, shall we make a `TODO:` comment so with can easily understand this part and upcoming action because IMO this section is not so easy to understand by just checking the code.

PR #662, opened 2024-08-07
fixed `onClose` function re-render issue

ankush-notifi:
fixed `onClose` function re-render issue


eric-notifi (commented):
Just want to double confirm if this PR is ready or not as looks like the `onClose` global method still not yet being used in any component

packages/notifi-react/lib/components/NavHeader.tsx
```diff
@@ -16,6 +17,8 @@ export type NavHeaderProps = {
 export const NavHeader: React.FC<PropsWithChildren<NavHeaderProps>> = (
   props,
 ) => {
+  const { globalCtas } = useGlobalStateContext();
```
eric-notifi, line 20: Rather than hardcode the globalCtas in this optional action component, we can pass the action related props in `NavHeader.tsx`
Then we can add add the respective props in the component which consume the `NaveHeader.tsx`. For example in `FtuAlertEdit.tsx`, we can do like below:
```ts
export type FtuAlertEditProps = {
  // ...others
  navHeaderRightCta?: {
    action: string; // We need to have a respective action in globalCtas,
    icon: IconType;
  };
};

// ...others

export const FtuAlertEdit: React.FC<FtuAlertEditProps> = (props) => {
  const { updateFtuStage } = useNotifiUserSettingContext();
  const [isLoading, setIsLoading] = React.useState(false);

  const onClick = async () => {
    setIsLoading(true);
    await updateFtuStage(FtuStage.Done);
    setIsLoading(false);
    props.onClickNext();
  };

  const { globalCtas } = useGlobalStateContext();

  return (
      // ...others
      <NavHeader
        // ...others
        rightCta={
          props.navHeaderRightCta &&
          globalCtas?.[props.navHeaderRightCta.action]
            ? {
                icon: props.navHeaderRightCta.icon,
                action: () => globalCtas[props.navHeaderRightCta!.action](),
              }
            : undefined
        }
      >
      // ... others
  )
};
```
IMO, This approach is preferable because we can grant more flexability to users (developers) so they will be able to do whatever action for the navBar topRightCta instead of being limited as `onClose`

packages/notifi-react/lib/components/Connect.tsx
```diff
@@ -87,6 +89,15 @@ export const Connect: React.FC<ConnectProps> = (props) => {
     <div className={clsx('notifi-connect', props.classNames?.container)}>
       <div className={clsx('notifi-connect-title', props.classNames?.title)}>
         {props.copy?.title ?? defaultCopy.connect.title}
+
```
eric-notifi, line 92: The same concept as mentioned on the comment above

packages/notifi-react/lib/components/NavHeader.tsx
```diff
@@ -33,10 +36,14 @@ export const NavHeader: React.FC<PropsWithChildren<NavHeaderProps>> = (
...
-        onClick={() => props.rightCta?.action()}
+        className="notifi-nav-header-right-cta"
+        onClick={() => props.rightCta?.action() ?? globalCtas?.onClose?.()}
       >
-        {props.rightCta?.icon ? <Icon type={props.rightCta?.icon} /> : null}
+        {props.rightCta?.icon ? (
+          <Icon type={props.rightCta?.icon} />
+        ) : globalCtas?.onClose ? (
+          <Icon type={'close'} />
+        ) : null}
```
eric-notifi, line 46: The same comment as previous one

PR #663, opened 2024-08-13
add unread filter toggle

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/src/components/HistoryList.tsx
+++ b/packages/notifi-dapp-example/src/components/HistoryList.tsx
@@ -22,14 +22,15 @@ export const HistoryList: React.FC<HistoryListProps> = ({
   historyDetailEntry,
 }) => {
   const [isLoadingMoreItems, setIsLoadingMoreItems] = useState(false);
-  const [isUnread, setIsUnread] = useState(false);
   const {
     isLoading: isLoadingHistoryItems,
     historyItems,
     hasNextPage,
     getHistoryItems,
     unreadCount,
-  } = useNotifiHistoryContext({ includeRead });
+    isIncludeRead,
+    setIsIncludeRead,
+  } = useNotifiHistoryContext();
 
   const mainRef = useRef<HTMLDivElement>(null);
 
@@ -78,8 +79,8 @@ export const HistoryList: React.FC<HistoryListProps> = ({
             <div className="text-notifi-text text-sm">Unreads</div>
             <Toggle
               disabled={false}
-              checked={isUnread}
-              onChange={() => setIsUnread(!isUnread)}
+              checked={!isIncludeRead}
+              onChange={() => setIsIncludeRead(!isIncludeRead)}
             />
           </div>
         </div>
--- a/packages/notifi-dapp-example/src/components/HistoryList.tsx
+++ b/packages/notifi-dapp-example/src/components/HistoryList.tsx
@@ -29,7 +29,7 @@ export const HistoryList: React.FC<HistoryListProps> = ({
     hasNextPage,
     getHistoryItems,
     unreadCount,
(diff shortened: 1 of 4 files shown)


packages/notifi-react/lib/context/NotifiHistoryContext.tsx
```diff
@@ -174,6 +179,14 @@ export const NotifiHistoryContextProvider: FC<
     }
   }, [frontendClientStatus, getHistoryItems, cardConfig]);
 
+  useEffect(() => {
+    // reload history items with unread filter toggle change
```
eric-notifi, line 183: I would suggest to change the comment because in `notifi-react` there is no `filter toggle change` concept.
We might want to make it more neutral like
```ts
// NOTE: refresh history while switch between the following modes: only showing unread (`isIncludeRead=false`) & showing all histories (`isIncludeRead=ture`)

File: packages/notifi-react/lib/context/NotifiHistoryContext.tsx [hunk 2]
```diff
@@ -303,6 +316,8 @@ export const NotifiHistoryContextProvider: FC<
         historyItems,
         unreadCount: unreadCount ?? 0,
         hasNextPage: cursorInfo.hasNextPage,
+        isIncludeRead,
```
eric-notifi, line 319: Do we have to expose `isIncludeRead`. For now, I did not see the necessary use case of using `isIncludeRead`, so I would un-expose this one to make the exposed context props as few as possible. 
Hence, if it become needed, we can expose when we need that. WDYT

packages/notifi-dapp-example/src/components/HistoryList.tsx
```diff
@@ -59,15 +62,27 @@ export const HistoryList: React.FC<HistoryListProps> = ({
...
+                {unreadCount} unread message{unreadCount > 1 ? 's' : ''}
+              </div>
+            ) : null}
+          </div>
+          <div className="flex flex-row gap-2">
+            <div className="text-notifi-text text-sm">Unreads</div>
+            <Toggle
+              disabled={false}
+              checked={!isIncludeRead}
+              onChange={() => setIsIncludeRead(!isIncludeRead)}
```
eric-notifi, line 83: Shall we use `setIsIncludeRead((prev)=>!prev)` to
1. avoid potential racing issue
2. get rid of the use of `isIncludeRead`, so we do not need to expose `isIncludeRead` property.

packages/notifi-react/lib/context/NotifiHistoryContext.tsx
```diff
@@ -35,6 +35,8 @@ export type NotifiHistoryContextType = {
   historyItems: HistoryItem[];
   unreadCount: number;
   hasNextPage: boolean;
+  isIncludeRead: boolean;
```
eric-notifi, line 38: As mention above, do you think it is a good idea to make this `isIncludeRead` as internal const instead of expose it out?

packages/notifi-dapp-example/src/components/HistoryList.tsx
marukohao: If we don't expose isIncludeRead, how should we set the value for checked, we may need to set up another state for this value and it looks like duplicated work

eric-notifi: as mention, we can use `setIsIncludeRead((prev)=>!prev)` to do so. the `prev` instance is the current state of `isIncludeRead`.

eric-notifi (approved):
A small nit shall we change the comment note from
```
// reload history items with unread filter toggle change
```
to
```
// NOTE: refresh history while switch between the following modes: only showing unread (`isIncludeRead=false`) & showing all histories (`isIncludeRead=ture`)
```

PR #678, opened 2024-09-07
Add pwa and service worker to example

kai-notifi:
Example for adding pwa and service worker into a nextjs app.

Steps
- npm install service worker (in this example, I'm using yalc to resolve those dependencies)
- Add service worker to public folder via webpack. This will copy the service worker into the public folder
      ```
  patterns: [
        {
          from: '.yalc/@notifi-network/notifi-web-push-service-worker/dist/notifi-service-worker.js',
          to: "../public/"
        },
```
- Initialize service worker in app (usually at root)
```
  React.useEffect(() => {
    initWebPushServiceWorker()
  }, []);
```
- Update service worker when Notifi account is created. This step also needs to be done when user has given permissions to pwa to publish notifications
 ```
 React.useEffect(() => {
    if (idToken) {
      console.log(process.env.NEXT_PUBLIC_TENANT_ID)
      console.log(process.env.NEXT_PUBLIC_ENV)
      const userAccount = (jwtDecode(idToken) as CustomJwtPayload).email
      tryCreateWebPushSubscription(userAccount, process.env.NEXT_PUBLIC_TENANT_ID ?? '', process.env.NEXT_PUBLIC_ENV ?? '');
    }
  }, [idToken]);
```


eric-notifi:
Since we already have an example repo, not sure if this PR is still needed? @kai-notifi

PR #687, opened 2024-09-17
Add analytics to service worker

kai-notifi:
Add user interaction analytics to service worker

### DIFF
--- a/packages/notifi-web-push-service-worker/lib/utils/service-worker.ts
+++ b/packages/notifi-web-push-service-worker/lib/utils/service-worker.ts
@@ -1,6 +1,7 @@
 import { NotifiEnvironment } from '@notifi-network/notifi-frontend-client';
 
 import {
+  NotifiNotificationData,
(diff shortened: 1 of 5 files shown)


packages/notifi-web-push-service-worker/lib/utils/service-worker.ts
```diff
@@ -97,6 +98,15 @@ export const isNotifiWebPushEventData = (
   const eventData = data as NotifiWebPushEventData;
   if (!('Subject' in eventData)) return false;
   if (!('Message' in eventData)) return false;
-  // if (!('EncryptedBlob' in eventData)) return false; // TODO: uncomment when BE is ready
+  if (!('EncryptedBlob' in eventData)) return false;
```
eric-notifi, line 101: I think we still not yet have the `'EncryptedBlob'` in the payload, which could break the app (I tested on prod):
Screenshot 2024-09-17 at 22 15 16

PR #702, opened 2024-10-11
fixed useEffect dependency

ankush-notifi:
Fixed: Connect Screen Stuck Issue
https://notifi.atlassian.net/browse/MVP-5582


eric-notifi (approved):
Test with `notifi-react-example-v2`, works fine. LGTM
(I would suggest to merge it after Suin PR being merged in case of any minor change needed)

PR #708, opened 2024-10-17
added on-close function on notifi-card

ankush-notifi:
https://notifi.atlassian.net/browse/MVP-5162

### DIFF
--- a/packages/notifi-react/lib/context/GlobalStateContext.tsx
+++ b/packages/notifi-react/lib/context/GlobalStateContext.tsx
@@ -18,7 +18,13 @@ export type GlobalError = {
   errorData?: string; // Allowing to pass custom string data to be parsed by JSON.parse
 };
 
-type GlobalCtas = Record<string, () => void>;
+export enum CtaTypeEnum {
+  onClose = 'onClose',
+}
+
+export type CtaType = `${CtaTypeEnum}`;
+
+export type GlobalCtas = Record<CtaType, () => void>;
 
 export type GlobalStateContextType = {
   globalLoading: GlobalLoading;
--- a/packages/notifi-react/lib/style/connect.css
+++ b/packages/notifi-react/lib/style/connect.css
@@ -25,6 +25,14 @@
   font-weight: 600;
   text-align: center;
   padding: 2.5rem 1rem 0.5rem 1rem;
+  position: relative;
+}
+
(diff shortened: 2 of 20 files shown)


packages/notifi-react/lib/context/GlobalStateContext.tsx
```diff
@@ -18,7 +18,13 @@ export type GlobalError = {
...
 };
 
-type GlobalCtas = Record<string, () => void>;
+export enum CtaTypeEnum {
+  onClose = 'onClose',
+}
+
+export type CtaType = `${CtaTypeEnum}`;
+
+export type GlobalCtas = Record<CtaType, () => void>;
```
eric-notifi, line 27: If the value is always `()=> void`, why don't we just simply set an Unioin type for Cta like below
```ts
export type Cta = 'onClose';

export type GlobalCtas = Record<Cta, () => void>;
```
This also reserves the flexibility if in the future like below:
```ts
export type Cta = 'onClose' | 'onXXX'

export type GlobalCtas = Record<Cta, Cta extends 'onClose' ? () => void : (sth: string) => void>;
```

packages/notifi-react/lib/style/connect.css
```diff
@@ -25,6 +25,14 @@
...
   text-align: center;
   padding: 2.5rem 1rem 0.5rem 1rem;
+  position: relative;
+}
+
+.notifi-connect-right-cta {
+  position: absolute;
+  top: 2.5rem;
+  right: 1rem;
+  cursor: pointer;
```
eric-notifi, line 35: If we already have the style defined for NavHeader component, do we still need to have one here?

eric-notifi (commented):
I think I am seeing some concerns.
Is there any possibility that we only want to turn on RightCta on certain view. Say, we do not want to render the close cta on Connect view but others
If so, we might also want to consider that case.
> Note: just an idea: Not sure whether we can adopt similar approach as the copy or classNames to allow to manipulate the ctas in specific view. (edited)

eric-notifi (commented):
We might not want to enable that on `notifi-react-example` because it is not officially on default SDK figma. 
Instead I think adding it on doc would be nice. (including inline comment on codebase as a simple guideline)

eric-notifi (approved):
LGTM

PR #720, opened 2024-11-01
MVP-5673-patch: prevent mis-emitting wsError event while from server refactor Type of getSubscriptionData

eric-notifi:

As title and check the detail on below comments (code-highlighting)

### DIFF
--- a/packages/notifi-graphql/lib/NotifiSubscriptionService.ts
+++ b/packages/notifi-graphql/lib/NotifiSubscriptionService.ts
@@ -201,49 +201,57 @@ export class NotifiSubscriptionService {
     });
 
     this._wsClient.on('connecting', () => {
-      console.log('WebSocket connecting'); // TODO: Remove before merge
       this.eventEmitter.emit('wsConnecting');
     });
 
     this._wsClient.on('connected', () => {
-      console.log('WebSocket connected'); // TODO: Remove before merge
-
       this.eventEmitter.emit('wsConnected', this._wsClient!);
     });
 
     this._wsClient.on('closed', (event) => {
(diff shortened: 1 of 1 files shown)


packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -201,49 +201,57 @@ export class NotifiSubscriptionService {
...
-      console.error('WebSocket error:', error); // TODO: Remove before merge
-      if (error instanceof Error) {
-        this.eventEmitter.emit('wsError', error);
+      if (error instanceof Error /*⬅ Client (browser) side error*/) {
+        return this.eventEmitter.emit('wsError', error);
       }
+      this.eventEmitter.emit('wsError', {
+        ...(error as Error),
+        message: 'NotifiEventEmitter: Server side or unknown error',
+      });
```
eric-notifi, line 222: Add this section to emit error while the error comes from server side environments which have different interface of standard Error

packages/notifi-graphql/lib/NotifiSubscriptionService.ts
```diff
@@ -201,49 +201,57 @@ export class NotifiSubscriptionService {
...
+
+  if (
+    typeof eventPayload === 'object' &&
+    eventPayload !== null &&
+    subscriptionEvent in eventPayload
   ) {
-    return (subscriptionData as Record<keyof NotifiSubscriptionEvents, any>)[
-      subscriptionEvent
-    ] as NotifiSubscriptionEvents[T][0];
+    return eventPayload[subscriptionEvent];
```
eric-notifi, line 254: Refactor `Type definition` to form the gql subscription payload to following GQL spec to make thing clear

PR #733, opened 2024-11-12
replaced subject with topic-name

ankush-notifi:
replaced subject with topic name in inbox message detail UI
https://notifi.atlassian.net/browse/MVP-5730

### DIFF
--- a/packages/notifi-react/lib/components/HistoryDetail.tsx
+++ b/packages/notifi-react/lib/components/HistoryDetail.tsx
@@ -12,7 +12,6 @@ type HistoryDetailProps = {
     container?: string;
     main?: string;
     subject?: string;
-    topicName?: string;
     timestamp?: string;
     message?: string;
   };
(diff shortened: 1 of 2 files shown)


packages/notifi-react/lib/components/HistoryDetail.tsx
```diff
@@ -58,11 +59,11 @@ export const HistoryDetail: React.FC<HistoryDetailProps> = (props) => {
...
         <div
           className={clsx(
-            'notifi-history-detail-subject',
-            props.classNames?.subject,
+            'notifi-history-detail-topic',
+            props.classNames?.topicName,
           )}
         >
-          {historyItem.subject}
+          {historyItem.topic}
```
eric-notifi, line 66: I guess the must needed change is this line. 
Do we have to change other lines? ex. changing classNames override & css class.

ankush-notifi: Yes, that’s correct. I wanted to make sure the props aligned accurately with the content, but I’ve reverted the other changes to keep it focused on the essential update and minimize impact.

eric-notifi:
LGTM, Just one last question, is this only for certain clients? or for all users. I just want to ensure we’re all aligned on the fact that existing users' UX will be affected when they update to the new version.

eric-notifi (approved):

PR #738, opened 2024-11-15
Use LoginWithWeb3 flow for Aptos & Movement

ghost:


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -655,24 +734,20 @@ export class NotifiFrontendClient {
   }
 
   async logIn(
-    signMessageParams: SignMessageParams,
+    signMessageParams: LoginParams,
```
eric-notifi, line 737: Shall we change the name of argument to `loginParams`, so people will not confuse it with the type of `SignMessageParams`

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -426,23 +430,27 @@ export type Uint8SignMessageFunction = (
...
   oidcProvider: Types.OidcProvider;
   jwt: string;
 };
+
 export type OidcSignInFunction = () => Promise<OidcCredentials>;
 
-export type AuthenticateResult = Signature | OidcCredentials;
-type Signature = string;
+export type AuthenticateResult =
+  | { signature: string; signedMessage: string }
```
eric-notifi, line 452: I think changing the interface here will break the login method becuase below in line746,  772 - 775, 807 - 810.

eric-notifi (approved):
LGTM 🚀🚀

PR #747, opened 2024-11-26
Enable LoginWithWeb3 for Cosmos chains

ghost:
- I tested that LoginWithWeb3 works for Cosmos signing methods. This PR will update our flows to use LoginWithWeb3 for all Cosmos chains.
- Alex informed me that we arent actively maintaining Evmos & Acala atm - marked them as such  in SDK for now.
- Organised the chains into distinct groups with helper functions & narrowing of types to enable easier development. There are various sanity checks in the types which should point developers to the necessary changes when adding a new chain...

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -31,6 +31,7 @@ import {
   CosmosBlockchain,
   EvmBlockchain,
   UnmaintainedBlockchain,
+  isAptosBlockchain,
   isCosmosBlockchain,
   isUsingAptosBlockchain,
   isUsingBtcBlockchain,
@@ -105,6 +106,10 @@ export type SignMessageParams =
   | Readonly<{
       walletBlockchain: 'OFF_CHAIN';
       signIn: OidcSignInFunction;
+    }>
+  | Readonly<{
+      walletBlockchain: 'INJECTIVE';
+      signMessage: Uint8SignMessageFunction;
     }>;
 
 /** NOTE:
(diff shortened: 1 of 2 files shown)


packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -855,63 +663,51 @@ export class NotifiFrontendClient {
...
+        await signMessageParams.signMessage(signMessageParams.message);
+      return { signature: signatureBase64, signedMessage };
+    } else if (isUsingAptosBlockchain(signMessageParams)) {
+      const { signatureHex, signedMessage } =
+        await signMessageParams.signMessage(
+          SIGNING_MESSAGE_WITHOUT_NONCE,
+          signMessageParams.nonce,
         );
+      return { signature: signatureHex, signedMessage };
+    } else if (isUsingEvmBlockchain(signMessageParams)) {
```
eric-notifi, line 686: In existing version, there are two cases to login `INJECTIVE` chain in `_authenticate` method, one under EVM, one under COSMOS. But this change will restrict users only using EVM. It might break the existing injective users.

packages/notifi-frontend-client/lib/client/blockchains.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// IGNORE BELOW HERE, this simpply serves to ensure all blockchains are covered
+// This will cause a type error if AllBlockchains differs from WalletBlockchain
+type BlockchainTypeCheck = [
+  Blockchain extends WalletBlockchain ? true : false,
+  WalletBlockchain extends Blockchain ? true : false,
+] extends [true, true]
+  ? true
+  : false;
```
eric-notifi, line 137: Wow .. This section is cool 👍 ty

eric-notifi:
Hmm ... not sure why the vercel deploy fails.
@nimesh-notifi , Could you help to check the deploy log as now only the account owner is able to see the deploy detail info. 🙏

eric-notifi (approved):
LGTM. 🔥
I believe this change will not be the breaking change for runtime. But I agree it is better to include it  into next major version (v4) because the the the reform of interface & type is huge.

eric-notifi:
@jamie-notifi I think we can merge it first if there is no other concern as it will not be immediately pumped to npm until we manually do so. We just want to keep in mine that the next version will be the major version `v4`

PR #784, opened 2025-02-07
MVP-6030: Update `notifi-react` to support `TenantConfigV2`

eric-notifi:
- Describe the purpose of the change

  - v2 functional version notifi-react using tenantConfigV2
  - Refactor tenantConfigV2 related modules


packages/notifi-frontend-client/lib/models/TenantConfig.ts
eric-notifi, line 1: @marukohao FYI, you can refer to this module for the v2 interface

PR #789, opened 2025-02-19
MG / Update Solana Hardware Wallet Package

mayuranganesathas:

### DIFF
--- a/packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
+++ b/packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
@@ -14,6 +14,7 @@ export type MemoProgramHardwareLoginPluginParams = Readonly<{
   walletPublicKey: string;
   connection: Connection;
   sendTransaction: WalletContextState['sendTransaction'];
+  signTransaction: WalletContextState['signTransaction'];
 }>;
 
 export class MemoProgramHardwareLoginPlugin implements HardwareLoginPlugin {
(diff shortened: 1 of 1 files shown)


packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
```diff
@@ -14,6 +14,7 @@ export type MemoProgramHardwareLoginPluginParams = Readonly<{
   walletPublicKey: string;
   connection: Connection;
   sendTransaction: WalletContextState['sendTransaction'];
+  signTransaction: WalletContextState['signTransaction'];
```
eric-notifi, line 17: Hmmm... well, directly adding a non-optional field and changing the underlaying implementation will cause the breaking change which we might want to avoid.

eric-notifi: I think this change might be the reason for the Vercel deployment CI failure.

packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
```diff
@@ -24,35 +25,33 @@ export class MemoProgramHardwareLoginPlugin implements HardwareLoginPlugin {
...
-    return signature;
+    if (!signTransaction) {
+      throw new Error('Wallet does not support transaction signing');
+    }
+    const signedTx = await signTransaction(txn);
+    const serializedTx = signedTx.serialize();
+
+    const base64Tx = Buffer.from(serializedTx).toString('base64');
+    return base64Tx;
   };
```
eric-notifi, line 56: This refactor looks like taking the same effect as the original one as both the new `signTransaction` and old `sendTransaction` methods are externally passed in methods (refer [here](https://github.com/notifi-network/notifi-sdk-ts/blob/a5a1b474011e2f002c82b78c8f47630c43a6baa5/packages/notifi-react-example-v2/src/context/NotifiContextWrapper.tsx#L85)).
Hope I did not miss anything 🙂

packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
```diff
@@ -24,35 +25,33 @@ export class MemoProgramHardwareLoginPlugin implements HardwareLoginPlugin {
   }
 
   sendMessage: (message: string) => Promise<string> = async (message) => {
```
eric-notifi, line 27: Not sure if we could add a new `signMessage` method to `HardWarePlugin`.  

We could implement `signMessage` using the new `signTransaction` parameter and mark the `sendXXX`-related parameters and methods as deprecated.  

This approach could:  
1. Improve readability  
2. Make the migration history clearer for users  
3. Avoid breaking changes

eric-notifi: By doing so, we can also clarify things by updating the README doc.

mayuranganesathas: gotcha will do

PR #793, opened 2025-02-21
MG / add new method and deprecate old

mayuranganesathas:

### DIFF
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -120,7 +120,9 @@ export type SignMessageParams =
 type SolanaWalletWithSignParams = Readonly<{
   signMessage: Uint8SignMessageFunction;
   hardwareLoginPlugin?: {
-    /** @deprecated No longer used. Will be removed in a future version. */
+    /**
+     * @deprecated Use signTransaction() instead. We no longer have to send a txn, and instead simply rely on the signed TX as we can verify this on Notifi Services.
+     */
     sendMessage: (message: string) => Promise<string>;
     signTransaction: (message: string) => Promise<string>;
   };
@@ -320,7 +322,7 @@ type FindSubscriptionCardParams = Omit<Types.FindTenantConfigInput, 'tenant'>;
 
 // Don't split this line into multiple lines due to some packagers or other build modules that
 // modify the string literal, which then causes authentication to fail due to different strings
-const SIGNING_MESSAGE_WITHOUT_NONCE = `Sign in with Notifi \n\n    No password needed or gas is needed. \n\n    Clicking "Approve" only means you have proved this wallet is owned by you! \n\n    This request will not trigger any transaction or cost any gas fees. \n\n    Use of our website and service is subject to our terms of service and privacy policy.`;
+const SIGNING_MESSAGE_WITHOUT_NONCE = `Sign in with Notifi \n\n    No password needed or gas is needed. \n\n    Clicking “Approve” only means you have proved this wallet is owned by you! \n\n    This request will not trigger any transaction or cost any gas fees. \n\n    Use of our website and service is subject to our terms of service and privacy policy.`;
 export const SIGNING_MESSAGE = `${SIGNING_MESSAGE_WITHOUT_NONCE} \n \n 'Nonce:' `;
 
 const CHAINS_WITH_LOGIN_WEB3 = [
--- a/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
+++ b/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
@@ -123,7 +123,7 @@ type SolanaWalletWithSignParams = Readonly<{
     /**
      * @deprecated Use signTransaction() instead. We no longer have to send a txn, and instead simply rely on the signed TX as we can verify this on Notifi Services.
      */
-    sendMessage: (message: string) => Promise<string>;
+    sendMessage?: (message: string) => Promise<string>;
     signTransaction: (message: string) => Promise<string>;
   };
 }> &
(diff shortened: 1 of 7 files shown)


mayuranganesathas:
@eric-notifi Do you propose publishing HW wallet indepdently, and then update the pkg json on the wrapper?

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -319,7 +320,7 @@ type FindSubscriptionCardParams = Omit<Types.FindTenantConfigInput, 'tenant'>;
 
 // Don't split this line into multiple lines due to some packagers or other build modules that
 // modify the string literal, which then causes authentication to fail due to different strings
-const SIGNING_MESSAGE_WITHOUT_NONCE = `Sign in with Notifi \n\n    No password needed or gas is needed. \n\n    Clicking “Approve” only means you have proved this wallet is owned by you! \n\n    This request will not trigger any transaction or cost any gas fees. \n\n    Use of our website and service is subject to our terms of service and privacy policy.`;
+const SIGNING_MESSAGE_WITHOUT_NONCE = `Sign in with Notifi \n\n    No password needed or gas is needed. \n\n    Clicking "Approve" only means you have proved this wallet is owned by you! \n\n    This request will not trigger any transaction or cost any gas fees. \n\n    Use of our website and service is subject to our terms of service and privacy policy.`;
```
eric-notifi, line 323: Are we supposed to change this string, looks risky. Could it break the existing sign methods??

eric-notifi (commented):
1. I guess we also want to update the notifi-react-example pkg as it also consumes the MemoProgramHardwareLoginPlugin. See [here](https://github.com/notifi-network/notifi-sdk-ts/blob/a5a1b474011e2f002c82b78c8f47630c43a6baa5/packages/notifi-react-example-v2/src/context/NotifiContextWrapper.tsx#L85)
2. I am not sure why, but looks like this chnage breaks the existing sign. You can try run `npx lerna --scope=@notifi-network/notifi-react-example-v2 run dev` and test trough the UI. it shows `SIGNATURE_DOES_NOT_MATCH` even when using ETH chain.

eric-notifi:
> RE: Do you propose publishing HW wallet indepdently, and then update the pkg json on the wrapper?

Well, I believe we don't need to publish separately. Lerna should manage the internal dependency tree within the monorepo for us.

packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
```diff
@@ -14,6 +15,7 @@ export type MemoProgramHardwareLoginPluginParams = Readonly<{
   walletPublicKey: string;
   connection: Connection;
   sendTransaction: WalletContextState['sendTransaction'];
```
eric-notifi, line 17: Do we want to make `sendTransaction` as optional?? 
If not, users will still have to input both sendTransaction & signTransaction when instantiate the class.

packages/notifi-react-example-v2/src/context/NotifiContextWrapper.tsx
```diff
@@ -82,7 +83,7 @@ export const NotifiContextWrapper: React.FC<PropsWithChildren> = ({
         solanaHardwareLoginPlugin = new MemoProgramHardwareLoginPlugin({
           walletPublicKey,
           connection,
-          sendTransaction: wallets[selectedWallet].signTransaction,
+          signTransaction: wallets[selectedWallet].signTransaction,
```
eric-notifi, line 86: As also mentioned above, if we do not make `sendTransaction` optional, we also have to put it here. Otherwise, typescript lint will be angry.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: i think this is whats breaking it. Let me add and test

packages/notifi-solana-hw-login/lib/plugin/MemoProgramHardwareLoginPlugin.ts
```diff
@@ -5,6 +5,7 @@ import {
   Transaction,
   TransactionInstruction,
 } from '@solana/web3.js';
+import base58 from 'bs58';
 
 export interface HardwareLoginPlugin {
   sendMessage: (message: string) => Promise<string>;
```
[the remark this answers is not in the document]
mayuranganesathas: done

packages/notifi-react-example-v2/src/context/NotifiContextWrapper.tsx
eric-notifi, line 86: Looks like we still have some type mismatch here. 
```
Type '(transaction: Transaction, connection: Connection) => Promise<string>' is not assignable to type '<T extends Transaction | VersionedTransaction>(transaction: T) => Promise<T>'.ts(2322)
index.d.ts(18, 5): The expected type comes from property 'signTransaction' which is declared here on type 'Readonly<{ walletPublicKey: string; connection: Connection; sendTransaction?: ((transaction: Transaction | VersionedTransaction, connection: Connection, options?: SendTransactionOptions | undefined) => Promise<...>) | undefined; signTransaction: (<T extends Transaction | VersionedTransaction>(transaction: T) => Prom...'
```
Screenshot 2025-02-24 at 11 28 52
I am not sure if this is the cause of deployment failure.

eric-notifi:
Hmmm ... looks like it still fails on vercel deployment. @nimesh-notifi could you help to take a look at the fail logs when you have a moment.
@mayuranganesathas , not sure if it only happens on my end, the hardware sign fails.
Screenshot 2025-02-24 at 12 55 03

mayuranganesathas:
> Hmmm ... looks like it still fails on vercel deployment. @nimesh-notifi could you help to take a look at the fail logs when you have a moment. @mayuranganesathas , not sure if it only happens on my end, the hardware sign fails. Screenshot 2025-02-24 at 12 55 03

Yeah - dev impl is using old path. New path confirms via TX through sol proxy

eric-notifi:
I see! That leaves just one remaining mystery: why the Vercel auto-deployment ended up in a failed state. 🥲

mayuranganesathas:
no longer using old deprecated path, utilizing web3 login path now

PR #798, opened 2025-02-28
Use Web3 Signin path for Solana and Solana Hardware Wallet

mayuranganesathas:


packages/notifi-wallet-provider/lib/types.ts
```diff
@@ -124,6 +124,9 @@ export class PhantomWallet implements NotifiWallet {
       transaction: Transaction,
       connection: Connection,
     ) => Promise<string>,
+    public signHardwareTransaction: (
+      transaction: Transaction,
+    ) => Promise<Transaction>,
```
eric-notifi, line 129: Just wondering do we have to have this method, or this is to replace `signTransaction` method??

packages/notifi-react-example-v2/.env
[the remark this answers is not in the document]
mayuranganesathas: mistake, fat fingers sorry

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -731,7 +773,7 @@ export class NotifiFrontendClient {
       case 'SOLANA': {
         const { walletPublicKey, tenantId } = this
           ._configuration as NotifiConfigWithPublicKey;
-        const signedMessage = `${SIGNING_MESSAGE}${walletPublicKey}${tenantId}${timestamp.toString()}`;
+        const signedMessage = `${SIGNING_MESSAGE}${signMessageParams.nonce}`;
```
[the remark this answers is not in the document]
mayuranganesathas: the normal sign was added to web3login (_prepareweb3 & loginwithweb3) so we can now use this method I believe. I was able to test and verify this for normal sign

packages/notifi-wallet-provider/lib/types.ts
mayuranganesathas: also this is supported by solana hardware wallet so should future proof against any new wallets enabling HW sign

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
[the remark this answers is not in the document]
eric-notifi: Sorry for not making it confusing. Let me clarify a bit. Instead of continuing to use the existing legacy switch statement, we might want to adopt the pattern Jamie created. For example, we could pull the logic out of the switch/case and move it up, using conditional statements instead. This would make it clearer that we're trying to migrate from the old pattern to the new one.

packages/notifi-react/lib/utils/frontendClient.ts
```diff
@@ -14,21 +15,27 @@ export const loginViaSolanaHardwareWallet = async (
...
+    signingPubkey: walletWithSignParams.walletPublicKey,
   });
 
-  if (logInResult?.completeLogInByTransaction === undefined) {
-    throw new Error('Log in failed');
+  if (logInResult?.completeLogInWithWeb3 === undefined) {
+    throw new Error('Log in via Web3 failed');
   }
-  return logInResult;
+  return logInResult.completeLogInWithWeb3.user;
```
eric-notifi, line 40: Just an idea—since we’re now using the same endpoint, would it be feasible to consolidate them within the login method?

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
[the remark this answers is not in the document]
mayuranganesathas: done

packages/notifi-react/lib/utils/frontendClient.ts
mayuranganesathas: lets move this into a sep ticket.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -594,6 +594,11 @@ export class NotifiFrontendClient {
       throw new Error('logInWith Web3 - Invalid signature - expected string');
     }
 
+    if (!authentication.signature) {
+      throw new Error(
+        'Signature expected! In order to continue, please sign the transaction to confirm your notifications.',
+      );
+    }
```
eric-notifi, line 602: 1. The new `if` statement looks like having similar effect of the one above (`if (!('signature' in authentication)) `). Shall we consolidate them?
2. A small nit: we might want to add the function name as prefix to the error msg. ex. logInWith Web3 - `

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -777,6 +782,10 @@ export class NotifiFrontendClient {
         const messageBuffer = new TextEncoder().encode(signedMessage);
 
         const signedBuffer = await signMessageParams.signMessage(messageBuffer);
+
+        if (!signedBuffer) {
+          throw Error('Signature not completed');
+        }
```
eric-notifi: Similarly, it might be more clear by adding the method name as prefix in error msg like `._authenticate - `

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -944,6 +953,8 @@ export class NotifiFrontendClient {
       ...input,
     });
 
+    await this._handleLogInResult(result.completeLogInWithWeb3.user);
+
```
eric-notifi, line 957: This implementation appears to be redundant with [this section](https://github.com/notifi-network/notifi-sdk-ts/blob/c00284d9b3d97a244de469759829b2db49752796/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts#L662). 

If retaining this code is necessary, we should add a comment or TODO annotation to clarify:
1. The purpose of this implementation
2. Which version is expected to be used in the future

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -771,24 +772,23 @@ export class NotifiFrontendClient {
...
+    } else if (isUsingSolanaBlockchain(signMessageParams)) {
+      const { walletPublicKey, tenantId } = this
+        ._configuration as NotifiConfigWithPublicKey;
+      const signedMessage = `${SIGNING_MESSAGE}${signMessageParams.nonce}`;
+      const messageBuffer = new TextEncoder().encode(signedMessage);
+
+      const signedBuffer = await signMessageParams.signMessage(messageBuffer);
+
+      if (!signedBuffer) {
+        throw Error('Signature not completed');
```
eric-notifi, line 784: Similarly, it might be more clear by adding the method name as prefix in error msg like `._authenticate - `

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: done done

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: done done, needed for sol hardware login ONLY

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: done done

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
```diff
@@ -953,6 +949,7 @@ export class NotifiFrontendClient {
       ...input,
     });
 
+    //This is to ensure that hardware wallet logins are given authentication.
     await this._handleLogInResult(result.completeLogInWithWeb3.user);
```
eric-notifi, line 953: Sorry, just a final question, are enventually use the _handleLogInResult here or the one in the outer layer (The one in login method) after we finally migrate all our chains to using loginWithWeb3. 
If we eventually want to remove either of them? shall we add a TODO on either of them just in case we do not forget about it

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
eric-notifi: Got it.  Thank you.  Sorry for the minor note, but should we add a TODO just in case we decide to make that change later? This way, we won't forget about it.

packages/notifi-react/lib/utils/frontendClient.ts
eric-notifi: It would be nice if we could also make a todo comment (And a ticket) here just in case we do not forget about it.

packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts
mayuranganesathas: Adding kind sir

eric-notifi (approved):

eric-notifi (commented):
Looks like we have conflict here. Could you help to take a look?

eric-notifi (changes requested):
Looks like we have conflict here. Could you help to take a look?

eric-notifi (approved):

PR #801, opened 2025-03-04
Add Sei, Swellchain, BOB

mayuranganesathas:

### DIFF
--- a/packages/notifi-frontend-client/lib/client/blockchains.ts
+++ b/packages/notifi-frontend-client/lib/client/blockchains.ts
@@ -36,6 +36,7 @@ export const EVM_BLOCKCHAINS = [
   'BERACHAIN',
   'OPTIMISM',
   'THE_ROOT_NETWORK',
+  'SEI',
   'SONIC',
   'ROME',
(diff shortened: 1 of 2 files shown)


mayuranganesathas:
This will break on the build because its not added in the BE yet. Just FYI.

packages/notifi-frontend-client/lib/client/blockchains.ts
```diff
@@ -25,6 +25,7 @@ export const EVM_BLOCKCHAINS = [
   'ARBITRUM',
   'AVALANCHE',
   'BINANCE',
+  'BOB',
```
eric-notifi, line 28: If the new chains are EVMs, we might also want to add them into [NotifiFrontnendClinet's login case](https://github.com/notifi-network/notifi-sdk-ts/blob/c00284d9b3d97a244de469759829b2db49752796/packages/notifi-frontend-client/lib/client/NotifiFrontendClient.ts#L625) as EVM chains have not yet been migrated to the web3 login system.

eric-notifi (approved):

PR #803, opened 2025-03-12
update in dapp-example new target flow

marukohao:
###  update the new target flow per Sam's feedbacks
- update the verify banner copy when no destination has been set up
Screenshot 2025-03-12 at 3 30 15 PM

- users can remove destinations even if there is only one destination has been set up
Screenshot 2025-03-12 at 3 30 45 PM

- if isContactInfoRequired is true, user have to set up at least one destination to enable the "next" button on Ftu step 1


packages/notifi-dapp-example/src/components/FtuTargetList.tsx
```diff
@@ -13,14 +14,23 @@ import { TargetList } from './TargetList';
 export type FtuTargetListProps = {
   contactInfo: CardConfigItemV1['contactInfo'];
   onClickNext?: () => void;
+  isContactInfoRequired?: boolean;
```
eric-notifi, line 17: Wondering if there is any concern not to directly grab the `isContactInfoRequired` property from `useNotifiTargetContext`?

packages/notifi-dapp-example/src/app/notifi/dashboard/page.tsx
```diff
@@ -68,10 +74,12 @@ export default function NotifiDashboard() {
             className="mt-5 mb-4"
           />
         </div>
-        {unVerifiedTargets.length > 0 && cardView === 'history' ? (
+        {(!hasValidTargetMoreThan(targetData, 0) && cardView === 'history') ||
+        (unVerifiedTargets.length > 0 && cardView === 'history') ? (
           <VerifyBanner
             setCardView={setCardView}
             isInHistoryDetail={isInHistoryDetail}
+            hasValidTargetMoreThanOne={hasValidTargetMoreThan(targetData, 0)}
```
eric-notifi, line 82: Why do we want to pass it in as a wrapper instead of directly use `hasValidTargetMoreThan` in the component?

eric-notifi (commented):
Also noticed that the cypress CI has error. 
Could you take a look?

packages/notifi-dapp-example/src/app/notifi/dashboard/page.tsx
marukohao: We pass `hasValidTargetMoreThan(targetData, 0)` as a prop to VerifyBanner instead of calling it directly within the component to avoid unnecessary re-fetching targetData. Since targetData is already available in dashboard

packages/notifi-dapp-example/src/components/FtuTargetList.tsx
eric-notifi: Ah sorry, `isContactInfoRequired` is under `cardConfig`. `cardConfig` is in `useNotifiTenantConfigContext`. Here is the example
```
const { cardConfig } = useNotifiTenantConfigContext();
```

eric-notifi (approved):

PR #805, opened 2025-03-13
feat: Add Solana wallet support to wallet provider

ankush-notifi:
https://notifi.atlassian.net/browse/MVP-6151


packages/notifi-wallet-provider/lib/context/NotifiWallets.tsx
```diff
@@ -134,6 +150,49 @@ const NotifiWallet: React.FC<PropsWithChildren> = ({ children }) => {
...
+    throwError,
+    selectWallet,
+    'solflare',
+  );
+  const ledger = useSolanaWallets(
+    setIsLoading,
+    throwError,
+    selectWallet,
+    'ledger',
+  );
```
eric-notifi, line 194: One concern:
Ledger seems to be more blockchain-agnostic, so it feels a bit odd to design it specifically for Solana. I’m wondering if we could make it more scalable by keeping it chain-agnostic instead?

packages/notifi-wallet-provider/lib/context/NotifiWallets.tsx
```diff
@@ -134,6 +150,49 @@ const NotifiWallet: React.FC<PropsWithChildren> = ({ children }) => {
 
   const phantom = usePhantom(setIsLoading, throwError, selectWallet);
 
+  const phantomSolana = useSolanaWallets(
+    setIsLoading,
+    throwError,
+    selectWallet,
+    'phantomSolana',
+  );
```
eric-notifi, line 158: Similarly, Phantom also supports multiple chains. I’m not sure if it’s possible to consolidate this with the existing Phantom section?

packages/notifi-wallet-provider/lib/utils/wallet.ts
```diff
@@ -12,4 +12,12 @@ export const walletsWebsiteLink: { [key in keyof Wallets]: string } = {
   zerion: 'https://zerion.io/',
   xion: 'https://xion.burnt.com/',
   phantom: 'https://phantom.app/download',
+  phantomSolana: 'https://phantom.app/download',
+  solflare: 'https://solflare.com/',
+  ledger: 'https://www.ledger.com/',
+  sollet: 'https://www.sollet.io/',
+  solletExtension:
```
eric-notifi, line 19: Since I noticed that we only assign the Extension suffix to the Sollet wallet, I’m wondering if this is intended to potentially avoid any naming conflicts or duplication issues?

PR #824, opened 2025-04-01
Add logout when disconnect in dapp example

marukohao:

### DIFF
--- a/packages/notifi-dapp-example/src/components/Disconnect.tsx
+++ b/packages/notifi-dapp-example/src/components/Disconnect.tsx
@@ -2,12 +2,14 @@
 
 import { Icon } from '@/assets/Icon';
 import { isEVMChain } from '@/utils/typeUtils';
+import { useNotifiFrontendClientContext } from '@notifi-network/notifi-react';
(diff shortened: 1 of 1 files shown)


packages/notifi-dapp-example/src/components/Disconnect.tsx
```diff
@@ -50,7 +52,10 @@ export default function Disconnect() {
           <div
             className={`cursor-pointer flex gap-x-5 px-4 text-notifi-tab-unselected-text w-56 py-2 bg-notifi-wallet-menu-card-bg border-t border-[#565A8D]`}
             onClick={() => {
-              if (keys) wallets[selectedWallet].disconnect();
+              if (keys) {
+                wallets[selectedWallet].disconnect();
+                logout();
+              }
```
eric-notifi, line 58: Per comment here: https://github.com/notifi-network/integration-full-screen-taiko/pull/10#discussion_r2023057859
But I tried the logout in FTU, it still does not clean up the jwt. And it shows the error below:
```
TypeError: logout is not a function
```
Screenshot 2025-04-03 at 22 01 15

PR #826, opened 2025-04-02
Fix tg/discord confirmation page redirect issue

marukohao:
Some wallet app browsers prevent to open link which with the deep link, for this case, we'll open the link on current page instead of opening the new window


eric-notifi (approved):
LGTM, There is also a refactored one on notifi-react just FYI:
https://github.com/notifi-network/notifi-sdk-ts/blob/16b84731764d45f2091da18573b045f91e51271b/packages/notifi-react/lib/hooks/useTargetListItem.tsx#L246

